PR #19, opened 2017-10-12
setup additions

cgosmeyer:
I've modified `setup.py` to install all the packages listed under "Dependencies" on http://grizli.readthedocs.io/en/master/grizli/install.html.  It ran smoothly for me, except when trying to install the package `reproject`.  It doesn't crash, just doesn't install it.  Not sure if it's just me...

To do the second half of the setup in http://grizli.readthedocs.io/en/master/grizli/install.html, I wrote `grizli/post_install.py` (might want to change the name) as an optional script that will create the GRIZLI directory and the iref, jref, CONF, and templates subdirectories, and will modify the `.bashrc` to include paths to GRIZLI, iref, and jref.  To make this work, I needed modify `grizli/utils.py` in the functions `fetch_hst_calib` and `fetch_default_calibs` to accept user-given paths to iref and jref.


gbrammer:
Thanks a lot, @cgosmeyer . Can you remove the requirement in post_install.py that "GRIZLI" be ${HOME}/GRIZLI?

cgosmeyer:
@gbrammer Originally I had a prompt where the user could type in the path where they wanted GRIZLI installed. Do you want something like that?

gbrammer:
@cgosmeyer: How about assuming the user has set the environment variable and taking it from `os.getenv`?

gbrammer:
Oh, now I notice that the script modifies ~/.bashrc.  I'd take that out and rely on the user to set the environment up themselves (for example, I wouldn't want to also require $iref to be within the Grizli path).

cgosmeyer:
@gbrammer What if I gave the user the option to either let the script modify the .bashrc or just print the lines they should add themselves?

gbrammer:
I'd rather not touch .bashrc at all.  In principle people could be running other shells as well.

cgosmeyer:
@gbrammer I see what you're saying!  Some may not like that, but I do, because I am lazy.  What if there were command line arguments that would give the user the option to create iref and jref directories, if they wanted to, and modify .bashrc, again just if they wanted to.  I assume most people would be in bash if using anaconda?  

```
>>> python post_install.py --iref /some/path --jref some/path --modify_bashrc True
```

gbrammer:
Sure, sounds good!

PR #44, opened 2020-05-13
Fixes

ivastar:
Minor changes to make it compatible with photutils >=0.7, astropy >= 0.4, astroquery >= 0.4. Edited the following notebooks and made sure the work:
Grizli Demo
Grizli-Pipeline
NewSpectrumFit


gbrammer:
Thanks, @ivastar.  I really messed this up when I went overboard the other day running `autopep8` on the whole repository, which totally broke the conflict resolution here.  I think I've got all of your commits in 29ca6a456f30e460c28d1a5fed46ee94a20219a6 other than your b9c7d4fd11d3bc5be09d5c06165ffac7288f1d75 and the updates to the notebooks.  Could you try pulling the latest and running the notebooks and just resend those if they work in a new PR?

gbrammer:
This has been sitting around for a while, but I'm going to close without merging because I think the new codebase has all of the fixes you implemented, with the exception of any changes to `detect_with_photutils`, which I never use or test.

PR #47, opened 2020-06-17
Adding Grizli for dummies to the Grizli repository

jkmatharu:
Grizli for dummies is an unofficial guide to using Grizli. Not sure how to create a link to it from the documentation page?


gbrammer:
Thanks a lot for this!  Can you convert to reStructuredText and put this in the `docs` directory along with a line in the `toctree` at the bottom of [docs/index.rst](https://github.com/gbrammer/grizli/blob/master/docs/index.rst) so that it gets rendered with the rest of the documentation?

gbrammer:
@jkmatharu Check the reorganization I made of the section headings to make sure that layout is how you had planned it.  Only the first section should have `========` so that it shows up as a single line in the main ToC.  Then subsections can be `------`, `~~~~~~~`, `^^^^^^^^`, etc.

gbrammer:
Scratch that @jkmatharu.  I bumped the headings back up since they can render OK as a separate tree in the ToC.  Please check the main repo to see if the main sections are now as you want them.

PR #61, opened 2022-04-05
JWST/NIRISS preprocessing updates

victoriastrait:
This pull request is to merge the changes from Victoria's work on making the grizli preprocessing routines work with NIRISS data. The major changes are as follows:

- autoscript.py
1. have *not* updated the "go" routine yet
2. added logic to collect *rate.fits files if isJWST==True 
3. passed isJWST to all relevant routines
4. in preprocess() change the instrument and detector back to NIRISS (had previously changed it to WFC3 to trick astrodrizzle)
5. in load_group_FLT() added logic to correctly initialize NIRISS group_flt objects

- prep.py
1. added logic in fresh_flt_file that points to the correct flat file for NIRISS data
6. adds functionality to do flat division for NIRISS images
7. In various places, added a search for "*rate.fits" files in addition to "*flc/t.fits"
8. call jwst_utils.img_with_wcs to update NIRISS wcs in process_direct_grism_visit
9. changed bits values for NIRISS data (right now it's happening in the "else" if an if statement, this could be improved)
10. updated a bunch of header keywords if 'isJWST==True' in process_direct_grism_visit, including changing the instrument name to WFC3, multiplying units by gain, and pointing to flat files. Also added flat field correction for direct images here. 
11. passed "isJWST" flag to all relevant routines
12. in visit_grism_sky() added information about background for NIRISS filters
13. in visit_grism_sky() added logic about the shape of NIRISS images using the isJWST flag (2048 for JWST, 1014 for HST)

- utils.py:
1. added an if statement for if "isJWST" flag==True in both parse_flt_files and parse_grism_associations. This allows visits to be parsed for NIRISS and HST images, depending on the isJWST flag, which must be passed as an input parameter in visit_prep_args. The biggest difference between how visits are parsed for NIRISS (vs HST) is that NIRISS includes a pupil parameter which is what is used to match a grism image to a direct image, whereas with HST a "best direct image" is chosen.


gbrammer:
This will be great, thanks @victoriastrait!  

The first thing I notice is that the files in `grizli/data/CONF` are included in the repo.  Can you zip these into a tarfile,  add relevant options to [utils.fetch_config_files](https://github.com/gbrammer/grizli/blob/master/grizli/utils.py#L5943) for downloading and unpacking them, and remove them from the repo?

gbrammer:
Can you also remove `grizli/data/iref` and `*npy` from `grizli/data/templates` from version control?

grizli/model.py
```diff
@@ -1606,9 +1612,10 @@ def __init__(self, sci=None, err=None, dq=None,
 
         self.wcs = None
 
-        if (instrument in ['NIRISS', 'NIRCAM']) & (~self.is_slice):
```
gbrammer, line 1609: why is this removed?

grizli/grismconf.py
```diff
@@ -506,6 +506,7 @@ def get_config_filename(instrume='WFC3', filter='F140W',
         String path of the configuration file.
 
     """
+
```
gbrammer, line 509: there is nothing different but this newline, can you just remove it so it doesn't show up as a diff?

grizli/data/auto_script_defaults.yml
```diff
@@ -10,7 +10,7 @@ filters: &filters
     [F410M, F467M, F547M, F550M, F621M, F689M, F763M, F845M, F200LP, F350LP,
      F435W, F438W, F439W, F450W, F475W, F475X, F555W, F569W, F600LP, F606W,
      F622W, F625W, F675W, F702W, F775W, F791W, F814W, F850LP, G800L, F098M,
-     F127M, F139M, F153M, F105W, F110W, F125W, F140W, F160W, G102, G141]
+     F127M, F139M, F153M, F105W, F110W, F125W, F140W, F160W, G102, G141, GR150C, F115W]
```
gbrammer, line 13: Add additional NIRISS and NIRCAm grisms and filters?

grizli/grismconf.py
victoriastrait: Fixed here: baf53da0ca6435830be3f5f34ef1e7dbf0e20011

victoriastrait:
> Can you also remove `grizli/data/iref` and `*npy` from `grizli/data/templates` from version control?

Done here: 39aebd67e15c8d6d1133f9581ae0d9e63964b6b3

grizli/model.py
victoriastrait: This was something I commented out while debugging and forgot to change back. Fixed here: 46eab444d0e74967c8198972f4ea6be2a638cf2f

grizli/data/auto_script_defaults.yml
victoriastrait: Added more filter names for NIRISS/NIRCam here: 4323a2902fda32fc117da85d0494d7b20fcd33e8

grizli/utils.py
```diff
@@ -1035,16 +1044,17 @@ def parse_visit_overlaps(visits, buffer=15.):
...
-
     return exposure_groups
 
 
 DIRECT_ORDER = {'G102': ['F105W', 'F110W', 'F098M', 'F125W', 'F140W', 'F160W', 'F127M', 'F139M', 'F153M', 'F132N', 'F130N', 'F128N', 'F126N', 'F164N', 'F167N'],
                 'G141': ['F140W', 'F160W', 'F125W', 'F105W', 'F110W', 'F098M', 'F127M', 'F139M', 'F153M', 'F132N', 'F130N', 'F128N', 'F126N', 'F164N', 'F167N'],
-                'G800L': ['F814W', 'F606W', 'F850LP', 'F775W', 'F435W', 'F105W', 'F110W', 'F098M', 'F125W', 'F140W', 'F160W', 'F127M', 'F139M', 'F153M', 'F132N', 'F130N', 'F128N', 'F126N', 'F164N', 'F167N']}
+                'G800L': ['F814W', 'F606W', 'F850LP', 'F775W', 'F435W', 'F105W', 'F110W', 'F098M', 'F125W', 'F140W', 'F160W', 'F127M', 'F139M', 'F153M', 'F132N', 'F130N', 'F128N', 'F126N', 'F164N', 'F167N'],
+                'GR150C': ['F115W', 'F150W', 'F200W'], 
+                'GR150R': ['F115W', 'F150W']}
```
gbrammer, line 1054: Add F200W

victoriastrait: Fixed: 17202984d46948a8449ff6b53f9368b82b64f5fc

gbrammer:
I removed `grizli/version.py` in a later commit on the main branch.  Can you also please remove it from version control here?

victoriastrait:
> I removed `grizli/version.py` in a later commit on the main branch. Can you also please remove it from version control here?

Done here: 598b4bc806a761a4931748b0e9683d92ecdd36da

gbrammer:
Can you wrap the `import jwst` bit at the top of [prep.py](https://github.com/victoriastrait/grizli/blob/JWST/grizli/prep.py#L25) within a try/except block so that we don't have to add the full jwst pipeline as an installation requirement?

e.g., 

```python
try:
    import jwst
    from jwst.pipeline import Detector1Pipeline
except ImportError:
    jwst = None
    print('`import jwst` failed so JWST processing will not work!')
```

gbrammer:
> Can you wrap the `import jwst` bit at the top of [prep.py](https://github.com/victoriastrait/grizli/blob/JWST/grizli/prep.py#L25) within a try/except block so that we don't have to add the full jwst pipeline as an installation requirement?
> 
> e.g.,
> 
> ```python
> try:
>     import jwst
>     from jwst.pipeline import Detector1Pipeline
> except ImportError:
>     jwst = None
>     print('`import jwst` failed so JWST processing will not work!')
> ```

OK, I realized that I could do this myself, so this little bit is done.

gbrammer:
As of https://github.com/gbrammer/grizli/pull/61/commits/c50ade25315bf495dabe2053174cdde35ced322c it seems like we've lost the templates again from `grizli/data`, which is why the checks failed.

victoriastrait:
> As of [c50ade2](https://github.com/gbrammer/grizli/commit/c50ade25315bf495dabe2053174cdde35ced322c) it seems like we've lost the templates again from `grizli/data`, which is why the checks failed.

These are put back now (without the *npy files). See here: e9bd2afd23fa0ea4cfaa3e0d68bfcaba05c258df

grizli/model.py
```diff
@@ -1431,7 +1434,10 @@ def __init__(self, sci=None, err=None, dq=None,
                 raise KeyError(msg)
 
             instrument = h['INSTRUME']
-            filter = utils.get_hst_filter(h, filter_only=True)
+            if instrument in ['NIRISS']:
```
gbrammer, line 1437: I think this isn't necessary now since `get_hst_filter(filter_only=True)` handles NIRISS and returns the `'FILTER'` keyword.

grizli/pipeline/default_params.py
```diff
@@ -14,6 +14,8 @@
 IR_M_FILTERS = ['F098M', 'F127M', 'F139M', 'F153M']
 IR_W_FILTERS = ['F105W', 'F110W', 'F125W', 'F140W', 'F160W']
 
+JW_FILTERS = ['F115W', 'F150W', 'F200W']
```
gbrammer, line 17: There are more filters, and can also add, e.g., `NIRCAM_LW_FILTERS = ...` and `NIRCAM_SW_FILTERS = ...`.   And eventually `MIRI_FILTERS`, but maybe that can wait until other MIRI logic is implemented.

gbrammer:
This is getting really close!  I think the last thing along some comments I added is to pull out `isJWST` from various places.

grizli/model.py
victoriastrait: Resolved here: 099acf4aecbcdc887c0980dccc530cedd6d2420c

grizli/multifit.py
```diff
@@ -3709,7 +3710,7 @@ def eval_trace_shift(shifts, self, indices, poly_order, lm, verbose, fit_with_ps
 
         return chi2/self.DoF
 
-    def drizzle_grisms_and_PAs(self, size=10, fcontam=0, flambda=False, scale=1, pixfrac=0.5, kernel='square', usewcs=False, tfit=None, diff=True, grism_list=['G800L', 'G102', 'G141', 'F090W', 'F115W', 'F150W', 'F200W', 'F356W', 'F410M', 'F444W'], mask_segmentation=True, reset_model=True, make_figure=True, fig_args=dict(mask_segmentation=True, average_only=False, scale_size=1, cmap='viridis_r'), **kwargs):
+    def drizzle_grisms_and_PAs(self, size=10, fcontam=0, flambda=False, scale=1, pixfrac=0.5, kernel='square', usewcs=False, tfit=None, diff=True, grism_list=['G800L', 'G102', 'GR150C', 'GR150R', 'G141', 'F090W', 'F115W', 'F150W', 'F200W', 'F356W', 'F410M', 'F444W'], mask_segmentation=True, reset_model=True, make_figure=True, fig_args=dict(mask_segmentation=True, average_only=False, scale_size=1, cmap='viridis_r'), **kwargs):
```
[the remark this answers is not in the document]
victoriastrait: Done: af9e6184ca376316be6b29ad22776320dd981fac

grizli/pipeline/default_params.py
victoriastrait: I've added imaging filters for NIRISS and NIRCAM here: 46d54879fc65c9a45098b391f5e02fa957523f18

But how do I list the NIRCam grisms?

grizli/pipeline/default_params.py
```diff
@@ -14,7 +14,10 @@
 IR_M_FILTERS = ['F098M', 'F127M', 'F139M', 'F153M']
 IR_W_FILTERS = ['F105W', 'F110W', 'F125W', 'F140W', 'F160W']
 
-JW_FILTERS = ['F115W', 'F150W', 'F200W']
+NIRISS_FILTERS = ['F090W', 'F115W', 'F150W', 'F200W', 'F277W', 'F356W', 'F444W', 'F140M', 'F158M', 'F380M', 'F430M', 'F480M']
+NIRCAM_LW_FILTERS = ['F322W2', 'F277W', 'F356W', 'F444W', 'F250M', 'F300M', 'F335M', 'F360M', 'F410M', 'F430M', 'F460M', 'F480M', 'F323N', 'F405N', 'F466N', 'F470N']
+NIRCAM_SW_FILTERS = ['F150W2', 'F070W', 'F090W', 'F115W', 'F150W', 'F200W', 'F140M', 'F162M', 'F182M', 'F210M', 'F164N', 'F187N', 'F212N']
```
gbrammer, line 19: This gets a little complicated with the filter names that are common to both NIRISS and NIRCam.  I guess `NIRCAM_SW_FILTERS + NIRCAM_LW_FILTERS`.  I think it's good to have them as separate lists, but `JW_FILTERS` should be just the unique values.

grizli/grismconf.py
gbrammer: Could it be that we lost this change with the other synchronization problems?  It still shows up as a diff here.

grizli/model.py
```diff
@@ -1645,7 +1649,7 @@ def unset_dq(self):
         """
 
         okbits_instrument = {'WFC3': 32+64+512,  # blob OK
-                             'NIRISS': 0,
+                             'NIRISS': 1+2+4+4096, #+4096+4100+18432+18436+1024+16384+1,
```
gbrammer, line 1652: Could you add a comment for what these bits mean, as far as you can tell?

grizli/model.py
```diff
@@ -2173,8 +2177,8 @@ def get_slice(self, slx=slice(480, 520), sly=slice(480, 520),
                               header=slice_header, wcs=slice_wcs,
                               photflam=self.photflam, photplam=self.photplam,
                               origin=slice_origin, instrument=self.instrument,
-                              filter=self.filter, pupil=self.pupil, 
-                              process_jwst_header=False)
+                              filter=self.filter,  pupil=self.pupil,
```
gbrammer, line 2180: please remove these extra spaces so they don't show up as a diff

grizli/model.py
```diff
@@ -3875,17 +3882,18 @@ def __init__(self, flt=None, beam=None, conf=None,
...
+                         seg_ids=None, min_sens=0.08, mask_resid=True, isJWST=False):
         """
         See parameter description for `~grizli.model.BeamCutout`.
         """
         # bad pixels or problems with uncertainties
-        self.mask = ((self.grism.data['DQ'] > 0) |
-                     (self.grism.data['ERR'] == 0) |
-                     (self.grism.data['SCI'] == 0))
+        self.mask = ((self.grism.data['DQ'] > 0) | 
+                    (self.grism.data['ERR'] == 0) |
```
gbrammer, line 3894: i think add back a space before `(` after the first line so that they don't show up as a change?

grizli/model.py
```diff
@@ -3998,13 +4006,12 @@ def init_from_input(self, flt, beam, conf=None, get_slice_header=True):
...
         if conf is None:
             conf = grismconf.load_grism_config(flt.conf_file)
-
         self.beam = GrismDisperser(id=beam.id, direct=beam.direct*1,
                            segmentation=beam.seg*1, origin=beam.origin,
                            pad=beam.pad, grow=beam.grow,
                            beam=beam.beam, conf=conf, xcenter=beam.xcenter,
                            ycenter=beam.ycenter, fwcpos=flt.grism.fwcpos,
-                           MW_EBV=flt.grism.MW_EBV)
+                           MW_EBV=flt.grism.MW_EBV) #
```
gbrammer, line 4014: remove

grizli/model.py
```diff
@@ -4079,6 +4086,7 @@ def load_fits(self, file, conf=None, direct_extn=1, grism_extn=2):
             direct_filter = self.direct.filter
 
         if conf is None:
+
```
gbrammer, line 4089: remove

grizli/model.py
```diff
@@ -4203,6 +4211,7 @@ def write_fits(self, root='beam_', overwrite=True, strip=False, include_model=Tr
         h0['MW_EBV'] = (self.grism.MW_EBV,
                          'Milky Way exctinction E(B-V)')
 
+
```
gbrammer, line 4214: remove

grizli/pipeline/auto_script.py
```diff
@@ -3,6 +3,7 @@
 """
 import os
 import inspect
+from this import d
```
gbrammer, line 6: ?

grizli/pipeline/auto_script.py
```diff
@@ -423,7 +424,7 @@ def go(root='j010311+131615',
                     parse_visits_args['max_dt'] = 4./24
                 else:
                     parse_visits_args['combine_same_pa'] = False
-                    parse_visits_args['max_dt'] = 1.
+                    parse_visits_args['max_dt'] = 1.
```
gbrammer, line 427: remove

grizli/pipeline/auto_script.py
```diff
@@ -2429,13 +2450,19 @@ def load_GroupFLT(field_root='j142724+334246', PREP_PATH='../Prep', force_ref=No
...
     info = utils.get_flt_info(files)
 
     g141 = info['FILTER'] == 'G141'
     g102 = info['FILTER'] == 'G102'
     g800l = info['FILTER'] == 'G800L'
+    gr150cf115w = (info['PUPIL'] == 'F115W') & (info['FILTER'] == 'GR150C')
+    gr150cf150w = (info['PUPIL'] == 'F150W') & (info['FILTER'] == 'GR150C')
+    gr150cf200w = (info['PUPIL'] == 'F200W') & (info['FILTER'] == 'GR150C')
+    gr150rf115w = (info['PUPIL'] == 'F115W') & (info['FILTER'] == 'GR150R')
+    gr150rf150w = (info['PUPIL'] == 'F150W') & (info['FILTER'] == 'GR150R')
```
gbrammer, line 2466: Missing the case of `gr150rf200w`

grizli/pipeline/auto_script.py
```diff
@@ -2518,6 +2545,192 @@ def load_GroupFLT(field_root='j142724+334246', PREP_PATH='../Prep', force_ref=No
 
         # del(grp_i)
 
+    if (gr150cf115w.sum() > 0) & ('GR150C' in gris_ref_filters):
```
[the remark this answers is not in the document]
gbrammer: It will get that much worse when we try to include all of the NIRCam permutations

grizli/pipeline/auto_script.py
```diff
@@ -2602,6 +2815,7 @@ def grism_prep(field_root='j142724+334246', PREP_PATH='../Prep', EXTRACT_PATH='.
                                 gris_ref_filters=gris_ref_filters, 
                                 files=files, split_by_grism=split_by_grism)
 
+
```
gbrammer, line 2818: remove

grizli/utils.py
```diff
@@ -1035,17 +1134,18 @@ def parse_visit_overlaps(visits, buffer=15.):
...
```
gbrammer, line 1144: F200W

grizli/prep.py
```diff
@@ -153,7 +187,7 @@ def fresh_flt_file(file, preserve_dq=False, path='../RAW/', verbose=True, extra_
         extra_msg = ' / flat: {0}'.format(flat_file)
 
         flat_im = pyfits.open(os.path.join(os.getenv('iref'), flat_file))
-        flat = flat_im['SCI'].data[5:-5, 5:-5]
+        flat = flat_im['SCI'].data [5:-5, 5:-5]
```
gbrammer, line 190: remove space. And these 5s are only appropriate for WFC3/IR.  Does JWST processing get to this point?

grizli/prep.py
```diff
@@ -350,7 +384,7 @@ def apply_persistence_mask(flt_file, path='../Persistence', dq_value=1024,
     flt = pyfits.open(flt_file, mode='update')
 
     pers_file = os.path.join(path,
-             os.path.basename(flt_file).replace('_flt.fits', '_persist.fits'))
+             os.path.basename(flt_file).replace('_flt.fits', '_persist.fits').replace('_rate.fits', '_persist.fits'))
```
gbrammer, line 387: this is probably OK, but we don't know what persistence products might (eventually) be available for JWST

grizli/prep.py
```diff
@@ -1824,7 +1858,7 @@ def make_SEP_catalog(root='',
             # tab = tab[~bad]
 
             # Correction for flux outside Kron aperture
-            tot_corr = get_wfc3ir_kron_tot_corr(tab, drz_filter, 
+            tot_corr = get_kron_tot_corr(tab, drz_filter,
```
gbrammer, line 1861: I agree to rename the function, but the function itself will have to be amended in some detail to work for the JWST instruments

grizli/prep.py
```diff
@@ -3146,23 +3181,29 @@ def process_direct_grism_visit(direct={},
...
+            if isJWST:
+                img = jwst_utils.img_with_wcs(file)
+                img.save(file)
+            else:
+                try:
+                    updatewcs.updatewcs(file, verbose=False, use_db=False)
+                except:
+                    updatewcs.updatewcs(file, verbose=False)
 
+
```
gbrammer, line 3193: remove these two?

grizli/prep.py
```diff
@@ -3146,23 +3181,29 @@ def process_direct_grism_visit(direct={},
...
 
+                
+        
         # ### Make ASN
         # if not isWFPC2:
         #     asn = asnutil.ASNTable(inlist=direct['files'], output=direct['product'])
         #     asn.create()
         #     asn.write()
 
+
```
gbrammer, line 3201: remove

grizli/prep.py
```diff
@@ -3202,13 +3252,14 @@ def process_direct_grism_visit(direct={},
         driz_cr_snr = '3.5 3.0'
         driz_cr_scale = '1.2 0.7'
     elif isWFPC2:
-        bits = 64+32
+        bits = 64+32 
         driz_cr_snr = '3.5 3.0'
         driz_cr_scale = '1.2 0.7'
     else:
-        bits = 576+256
+        bits  = 576+256
```
gbrammer, line 3259: remove

grizli/prep.py
```diff
@@ -3202,13 +3252,14 @@ def process_direct_grism_visit(direct={},
...
         driz_cr_snr = '8.0 5.0'
         driz_cr_scale = '2.5 0.7'
+
```
gbrammer, line 3262: remove

grizli/prep.py
```diff
@@ -3271,26 +3410,33 @@ def process_direct_grism_visit(direct={},
...
-                         clean=True, context=False, preserve=False,
+                         clean=True, context=False, preserve=False, 
                          skysub=True, driz_separate=True, driz_sep_wcs=True,
                          median=True, blot=True, driz_cr=True,
                          driz_cr_snr=driz_cr_snr_first,
                          driz_cr_scale=driz_cr_scale_first,
                          driz_cr_corr=False, driz_combine=True,
                          final_bits=bits, coeffs=True, build=False,
-                         final_wht_type='IVM', **drizzle_params)
+                         final_wht_type='IVM', gain='-99', rdnoise='-99', **drizzle_params)
```
gbrammer, line 3431: we might need to be careful with these.  Also the "ERR" extensions of JWST seem to be split into different components for read noise and Poisson.

grizli/prep.py
```diff
@@ -3499,8 +3647,8 @@ def process_direct_grism_visit(direct={},
             table_to_radec(cat[clip][so], '{0}.cat.radec'.format(direct['product']))
 
         if (fix_stars) & (not isACS) & (not isWFPC2):
-            fix_star_centers(root=direct['product'], drizzle=False, 
-                             mag_lim=19.5)
+            fix_star_centers(root=direct['product'], drizzle=False, isJWST=isJWST,
```
gbrammer, line 3650: Let's not do this at all for JWST for now

grizli/prep.py
```diff
@@ -3700,9 +3855,14 @@ def tweak_align(direct_group={}, grism_group={}, max_dist=1., n_min=10, key=' ',
         return True
 
     # Redrizzle
-    bits = 576
-    driz_cr_snr = '8.0 5.0'
-    driz_cr_scale = '2.5 0.7'
+    if isJWST:
+        bits = 1
```
gbrammer, line 3859: Again, the default values above are for WFC3/IR, and I don't think they should need to be that different for JWST.  But who knows for now, we'll have to adjust based on the real data.

grizli/prep.py
```diff
@@ -4480,15 +4688,20 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
     # Build combined arrays
     if isACS:
         bits = 64+32
+    elif isJWST:
+        bits = 1+4+6+32768+16777200+1049600+26232800+9438210+9438220
```
gbrammer, line 4692: These are going to be trouble for `int16`

grizli/prep.py
```diff
@@ -4480,15 +4688,20 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
...
     else:
-        bits = 576
+        bits =  576
```
gbrammer, line 4694: remove

grizli/prep.py
```diff
@@ -4480,15 +4688,20 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
...
 
     for i in range(Nexp):
         flt = pyfits.open(grism['files'][i])
         dq = utils.unset_dq_bits(flt['DQ', ext].data, okbits=bits)
         dq_mask = dq == 0
 
         # Data
+        print(ext)
```
gbrammer, line 4702: remove these debugging lines

grizli/prep.py
```diff
@@ -4716,14 +4932,18 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
                         label=grism['files'][j].split('_fl')[0])
 
         # result
-        fp = open(file.replace('_flt.fits', '_column.dat'), 'wb')
+        fp = open(file.replace('_flt.fits', '_column.dat').replace('_rate.fits', '_column.dat'), 'wb')
```
gbrammer, line 4935: Are we sure that the JWST grism exposures are being rotated correctly in this `prep` processing so that the column subtraction is truly columns parallel to the dispersion axis as expected?

grizli/prep.py
```diff
@@ -4716,14 +4932,18 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
...
         fp.write(b'# column obs_resid ok resid uncertainty\n')
         np.savetxt(fp, np.array([xmsk, yres, yok*1, y_pred-bg_sky, gp_sigma]).T, fmt='%.5f')
         fp.close()
 
         if apply:
             # Subtract the column average in 2D & log header keywords
-            gp_res = np.dot(y_pred[:, None]-bg_sky, np.ones((1014, 1)).T).T
+            if isJWST:
```
gbrammer, line 4942: Can get the required shape from `im_shape` and avoid the `if/else` statement here

grizli/prep.py
```diff
@@ -5387,7 +5612,7 @@ def drizzle_overlaps(exposure_groups, parse_visits=False, check_overlaps=True, m
             if isACS | isWFPC2:
                 bits = 64+32
             else:
-                bits = 576
+                bits  =  576
```
gbrammer, line 5615: remove extra spaces

grizli/prep.py
```diff
@@ -5571,7 +5796,7 @@ def extract_fits_log(file='idk106ckq_flt.fits', get_dq=True):
     log['chips'] = []
 
     if get_dq:
-        idx = np.arange(1014**2, dtype=np.int32).reshape((1014, 1014))
+        idx = np.arange(2048**2, dtype=np.int32).reshape((2048, 2048))
```
gbrammer, line 5799: This doesn't seem to be in a JWST logic block?  Should the appropriate size be determined somewhere above?

gbrammer:
@victoriastrait Lots of fairly small comments above, but one larger one was that I didn't notice in the `prep.py` updates for JWST if the JWST slitless exposures were being rotated as necessary for the column-average correction (and then rotated back).

grizli/pipeline/auto_script.py
```diff
@@ -1389,20 +1398,20 @@ def parse_visits(field_root='', RAW_PATH='../RAW', use_visit=True, combine_same_
 
             visits.append(direct)
 
-        all_groups = utils.parse_grism_associations(visits)
-        #np.save('{0}_visits.npy'.format(field_root), 
-        #        [visits, all_groups, info])
+        all_groups = utils.parse_grism_associations(visits, info, isJWST-isJWST)
```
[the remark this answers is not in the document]
victoriastrait: It will get taken out but fixed here anyway: e8d2bb869c605c4fbc6542dc4a493fbb8f8eb95a

grizli/pipeline/auto_script.py
victoriastrait: Fixed: fef84a382c3aa36b4261c5189d21ac09e9d5261d

grizli/pipeline/auto_script.py
victoriastrait: Fixed f7d2f0d8eca3bc9342cb6b434023dc7bd67bf6a2

grizli/pipeline/auto_script.py
```diff
@@ -3602,20 +3817,24 @@ def drizzle_overlaps(field_root, filters=['F098M', 'F105W', 'F110W', 'F125W', 'F
         wfc3ir['reference_wcs'] = ref_wcs
 
     filter_groups = {}
-    for visit in visits:
 
+    for visit in visits:
+        print(visit)
```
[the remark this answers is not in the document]
victoriastrait: Fixed: 970e89d0ba5f2baf248a3bbf4ca280b194ae53c6

grizli/utils.py
victoriastrait: Fixed: a24917d398ba1a027dd3d30d27d9aa058c04c25f

grizli/prep.py
```diff
@@ -1,6 +1,7 @@
 """
 Align direct images & make mosaics
 """
+from codecs import ignore_errors
```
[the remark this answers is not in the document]
victoriastrait: I don't know where that came from, I've never used this import before. Anyway, it's fixed: f6c72294fdfb8d03bce5c66176a24d10ca8d41c6

grizli/prep.py
victoriastrait: Fixed: 94713d82b411007ad54fb5022e913b17e1fd3287

grizli/prep.py
victoriastrait: This was actually you, I didn't rename this function

grizli/prep.py
victoriastrait: Fixed: 0a560a3e05460295eca1febbef24e4c358ef396c

grizli/prep.py
victoriastrait: Fixed: 0a560a3e05460295eca1febbef24e4c358ef396c

grizli/prep.py
```diff
@@ -3202,13 +3252,14 @@ def process_direct_grism_visit(direct={},
         driz_cr_snr = '3.5 3.0'
         driz_cr_scale = '1.2 0.7'
     elif isWFPC2:
-        bits = 64+32
+        bits = 64+32
```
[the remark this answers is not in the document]
victoriastrait: Fixed: 2b34d10b498e22e65dc6b60680c99db12c11beb7

grizli/prep.py
victoriastrait: Fixed: eac694a7a7e4364ca853ff49a374fe1af7c6d60b

grizli/prep.py
victoriastrait: Fixed: 92a1ab43521be8217bc96d744e472d318ea1b7fa

grizli/prep.py
victoriastrait: Fixed: 5523235dd71bb1b77d7043eae1569cba8b2362cd

grizli/prep.py
```diff
@@ -3514,19 +3662,25 @@ def process_direct_grism_visit(direct={},
 
     # First drizzle to flag CRs
     gris_cr_corr = len(grism['files']) > 1
-
+    if isJWST:
+        driz_cr_scale = '1.2 0.7'
+        driz_cr_snr = '35.0 15.0'
```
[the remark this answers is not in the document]
victoriastrait: This was something I was trying when figuring out the bits info for JWST and forgot to change back. I think you're right so I've removed the instances of this: 09c29642f78b0417e5251f73af0c9e8fc47b53a1

grizli/prep.py
victoriastrait: Fixed: 09c29642f78b0417e5251f73af0c9e8fc47b53a1

grizli/prep.py
victoriastrait: Fixed: 8bbb7d929cd70ce8ade796d8e2818cbaefbc7634

grizli/prep.py
victoriastrait: Fixed: 43d839a5640310eb4973d16244075d561c06c300

grizli/prep.py
victoriastrait: Fixed: 49636841cd102e75a1c0f96f17c1849bcdb344c2

grizli/prep.py
victoriastrait: I added the if statement for now, but I hope to get rid of "isJWST" and the hard coded sizes eventually: e7d4f097feecc58bf9d182979f00c7a13b083fe1

victoriastrait:
> There is a keyword `isJWST` that appears in a number of places in [c8d3bbf](https://github.com/gbrammer/grizli/commit/c8d3bbfc299927777a7796f68d370063c8e4a5c0) and also in `prep.py`. I think it would be safer in general and also easier to maintain the code if the relevant functions figure out what `isJWST` should be based on the other information they have been passed, e.g, data filenames or filters.

I've begun changing this. In the following commit I've added an automatic check for jwst data in parse_visits: 62da31e79dfad628fdbb69e3283facfca72739f7 Because of the nature of parse_visits, I only check for rate files here. In other functions, I will check the 'TELESCOP' header keyword for this

victoriastrait:
> > There is a keyword `isJWST` that appears in a number of places in [c8d3bbf](https://github.com/gbrammer/grizli/commit/c8d3bbfc299927777a7796f68d370063c8e4a5c0) and also in `prep.py`. I think it would be safer in general and also easier to maintain the code if the relevant functions figure out what `isJWST` should be based on the other information they have been passed, e.g, data filenames or filters.
> 
> I've begun changing this. In the following commit I've added an automatic check for jwst data in parse_visits: [62da31e](https://github.com/gbrammer/grizli/commit/62da31e79dfad628fdbb69e3283facfca72739f7) Because of the nature of parse_visits, I only check for rate files here. In other functions, I will check the 'TELESCOP' header keyword for this

Added some more automatic checks for JWST data: d76f72515ce45b1fdf0c1203ff926fcd485f29b4

grizli/prep.py
```diff
@@ -5595,3 +5808,19 @@ def extract_fits_log(file='idk106ckq_flt.fits', get_dq=True):
...
+    Check if a file or list of files is JWST image. 
+    Parameters:
+    - infile: fits file
+    
+    Returns:
+    - True or False
+    '''
+
+    hdu = pyfits.open(infile)
+    isJWST = (hdu[0].header['TELESCOP'] == 'JWST')
```
gbrammer, line 5824: This is now breaking in the test.  First add a check to see if the 'TELESCOP' keyword exists, and I guess assume `isJWST=False` if it doesn't.  

Also please try to make the formatting of the docstring look like that of other functions, e.g., 
```python
def column_values_in_list(col, test_list):
    """Test if column elements "in" an iterable (e.g., a list of strings)

    Parameters
    ----------
    col : `astropy.table.Column` or other iterable
        Group of entries to test

    test_list : iterable
        List of values to search

    Returns
    -------
    test : bool array
        Simple test:
            >>> [c_i in test_list for c_i in col]
    """
    test = np.array([c_i in test_list for c_i in col])
    return test
```

victoriastrait: Fixed both here (sorry!): 91838945f5bf1813894b49f9807a84bfb0e3118f

grizli/pipeline/auto_script.py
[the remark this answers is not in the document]
victoriastrait: Done here for the JWST filters, and I think it should work for the HST combinations as well but I haven't tested it. I will uncomment out the HST code block after testing with HST data: ec207f53c794f047bc9deafc952f2dd260793d31

grizli/pipeline/auto_script.py
victoriastrait: Fixed: 53d464fd16b06f72dc0a3cd401cb33e3701f08fd

grizli/pipeline/auto_script.py
victoriastrait: No idea what this is. Removed here: 059feb865d4304f355e85be3990691c20593edc5

grizli/model.py
victoriastrait: Done here: fa05f6b07e4aae965395ce2b14ffc95630d5f483

grizli/model.py
victoriastrait: Fixed here: 0da92a5224735dd7df6e8adf6827be5a5fe2ddab

grizli/model.py
victoriastrait: Done here: d11381b8f3e969cdc78a22340b5d3f9b85ae9e31

grizli/model.py
victoriastrait: Fixed here (I think?): 43c6ccd08f26c47b8448ef796eaddc0419d47be4

grizli/model.py
victoriastrait: Done here: 968c95f3cb1f439ed783c26ff4dd4dde4bcd9d48

grizli/pipeline/default_params.py
victoriastrait: Fixed here: 1b85be942c041dbd750a7037d0f4e5087e9190ff

grizli/pipeline/default_params.py
victoriastrait: Added MIRI filters here: e59c0b7c85120da783ef7348939772292e0a233f

grizli/prep.py
victoriastrait: fixed this and a few other instances here: 60d97b9d92fc01949ea01c96cba3c660b9ba8478

grizli/pipeline/auto_script.py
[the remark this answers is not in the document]
victoriastrait: Works for HST filters now: e2e51fca2829cf01e62df79f17fb26ff0029b6e0

grizli/tests/test_jwst.py
```diff
@@ -0,0 +1,31 @@
...
+
+class JWSTFittingTools(unittest.TestCase):
+    
+    def test_multibeam(self):
+        """
+        Can we initialize a multibeam file?
+        """
+        path = os.path.dirname(utils.__file__)
+        print(path)
+        beams_file = '/Users/victoriastrait/Desktop/grizli_test_data/jw01324001001_01243.beams.fits'
```
gbrammer, line 20: This will fail in the online tests. You could make a test to quit gracefully if not os.path.exists(beams_file), or if the file isn't too big you could put it in grizli/tests/data in the repo. If you do the latter, it would be good to call it something like niriss_jw01....beams.fits.

grizli/jwst_utils.py
```diff
@@ -122,11 +122,15 @@ def img_with_wcs(input):
...
     # HDUList -> jwst.datamodels.ImageModel
-    
+
     # Generate WCS as image
+    if not isinstance(input, pyfits.HDUList):
+        input = pyfits.open(input)
     if isinstance(input, pyfits.HDUList):
         if input[0].header['INSTRUME'] == 'NIRISS':
             if input[0].header['FILTER'].startswith('GR'):
+                input[0].header['FILTER_INFO'] = input[0].header['FILTER'] # just to save it so we can change it back at the end
```
gbrammer, line 132: FITS header keywords generally need to be 8 characters or less, though I'm not too careful with that elsewhere.  Maybe `OFILTER` and `OEXPTYPE`? (those are "Oh's", not zeros)

grizli/multifit.py
```diff
@@ -3709,7 +3716,7 @@ def eval_trace_shift(shifts, self, indices, poly_order, lm, verbose, fit_with_ps
 
         return chi2/self.DoF
 
-    def drizzle_grisms_and_PAs(self, size=10, fcontam=0, flambda=False, scale=1, pixfrac=0.5, kernel='square', usewcs=False, tfit=None, diff=True, grism_list=['G800L', 'G102', 'G141', 'F090W', 'F115W', 'F150W', 'F200W', 'F356W', 'F410M', 'F444W'], mask_segmentation=True, reset_model=True, make_figure=True, fig_args=dict(mask_segmentation=True, average_only=False, scale_size=1, cmap='viridis_r'), **kwargs):
+    def drizzle_grisms_and_PAs(self, size=10, fcontam=0, flambda=False, scale=1, pixfrac=0.5, kernel='square', usewcs=False, tfit=None, diff=True, grism_list=['G800L', 'G102', 'G141', 'GR150C', 'GR150R', 'F090W', 'F115W', 'F150W', 'F200W', 'F356W', 'F410M', 'F444W'], mask_segmentation=True, reset_model=True, make_figure=True, fig_args=dict(mask_segmentation=True, average_only=False, scale_size=1, cmap='viridis_r'), **kwargs):
```
gbrammer, line 3719: There's something inconsistent here with mixing JWST dispersers and bandpass filters.  Maybe we should call the dispersive element `dispname` throughout, which will be `FILTER` for HST, `PUPIL-FILTER` for NIRISS and `FILTER-PUPIL` for NIRCam, with a function in `utils` that generates this from a passed `Header` or `ImageHDU`.

grizli/pipeline/default_params.py
```diff
@@ -14,6 +14,12 @@
 IR_M_FILTERS = ['F098M', 'F127M', 'F139M', 'F153M']
 IR_W_FILTERS = ['F105W', 'F110W', 'F125W', 'F140W', 'F160W']
 
+NIRISS_FILTERS = ['F090W', 'F115W', 'F150W', 'F200W', 'F277W', 'F356W', 'F444W', 'F140M', 'F158M', 'F380M', 'F430M', 'F480M']
```
gbrammer, line 17: Please wrap at <79 characters here and below for PEP8, perhaps wrapping at logical breaks like between F200W and F277W for NIRISS (F277W and redder are in the `FILTER` wheel while the bluer filters are in the `PUPIL` wheel....) and between the `W`, `M` and `N` filters for NIRCam.

grizli/pipeline/photoz.py
```diff
@@ -271,7 +271,7 @@ def eazy_photoz(root, force=False, object_only=True, apply_background=True, aper
         cat = utils.read_catalog('{0}_phot_apcorr.fits'.format(root))
         return self, cat, zout
 
-    trans = {'f098m': 201, 'f105w': 202, 'f110w': 241, 'f125w': 203, 'f140w': 204, 'f160w': 205, 'f435w': 233, 'f475w': 234, 'f555w': 235, 'f606w': 236, 'f625w': 237, 'f775w': 238, 'f814w': 239, 'f850lp': 240, 'f702w': 15, 'f600lpu': 243, 'f225wu': 207, 'f275wu': 208, 'f336wu': 209, 'f350lpu': 339, 'f438wu': 211, 'f475wu': 212, 'f475xu': 242, 'f555wu': 213, 'f606wu': 214, 'f625wu': 215, 'f775wu': 216, 'f814wu': 217, 'f390wu': 210, 'ch1': 18, 'ch2': 19, 'f336w':209, 'f350lp':339}
+    trans = {'f098m': 201, 'f105w': 202, 'f110w': 241, 'f125w': 203, 'f140w': 204, 'f160w': 205, 'f435w': 233, 'f475w': 234, 'f555w': 235, 'f606w': 236, 'f625w': 237, 'f775w': 238, 'f814w': 239, 'f850lp': 240, 'f702w': 15, 'f600lpu': 243, 'f225wu': 207, 'f275wu': 208, 'f336wu': 209, 'f350lpu': 339, 'f438wu': 211, 'f475wu': 212, 'f475xu': 242, 'f555wu': 213, 'f606wu': 214, 'f625wu': 215, 'f775wu': 216, 'f814wu': 217, 'f390wu': 210, 'ch1': 18, 'ch2': 19, 'f336w':209, 'f350lp':339, 'f115w': 309, 'f150w': 310, 'f200w': 311}
```
gbrammer, line 274: Eventually add all other filters...

grizli/prep.py
```diff
@@ -6,6 +6,7 @@
 
 from collections import OrderedDict
 import glob
+from tabnanny import check
```
gbrammer, line 9: what's this?

grizli/prep.py
```diff
@@ -145,6 +155,30 @@ def fresh_flt_file(file, preserve_dq=False, path='../RAW/', verbose=True, extra_
                                calib_types=calib_types,
                                verbose=False)
 
+    if filter in ['GR150C', 'GR150R']: 
+        if (head['FILTER'] == 'GR150C') & (head['PUPIL'] == 'F115W'):
```
gbrammer, line 159: Getting these directly from CDBS can wait, but this could be cleaned up into a dictionary for now,  like 
```pythono
niriss_flats = {'GR150R': 
                         {'F115W': 'jwst_niriss_flat_0202.fits', 
                          'F150W': 'jwst_niriss_flat_0187.fits', 
                          ...}, 
                'GR150C': ...
                }

flat_file = niriss_flats[FILTER][PUPIL]
```

Note that `jref` is usually for HST ACS (`jref` for WFC3), but OK to use as shorthand for now until it's fully implemented with CRDS.

grizli/prep.py
```diff
@@ -430,7 +464,7 @@ def apply_region_mask(flt_file, dq_value=1024, verbose=True):
     """
     import pyregion
 
-    mask_files = glob.glob(flt_file.replace('_flt.fits', '.*.mask.reg').replace('_flc.fits', '.*.mask.reg').replace('_c0m.fits', '.*.mask.reg').replace('_c0f.fits', '.*.mask.reg'))
+    mask_files = glob.glob(flt_file.replace('_flt.fits', '.*.mask.reg').replace('_flc.fits', '.*.mask.reg').replace('_c0m.fits', '.*.mask.reg').replace('_c0f.fits', '.*.mask.reg').replace('_rate.fits','.*.mask.reg'))
```
gbrammer, line 467: I'm embarrassed by this line.  Let's do the following to strip out any extension beyond the last `_` character.

```python
# e.g., iabc01defq_flt.fits > iabc01defq.*.mask.reg
mask_files = glob.glob('_'.join(flt_file.split('_')[:-1]) + '.*.mask.reg')
```

grizli/prep.py
```diff
@@ -3146,11 +3180,17 @@ def process_direct_grism_visit(direct={},
...
-            except:
-                updatewcs.updatewcs(file, verbose=False)
+            isJWST = check_isJWST(file)
+            if isJWST:
+                img = jwst_utils.img_with_wcs(file)
+                img.save(file)
+            else:
+                try:
+                    updatewcs.updatewcs(file, verbose=False, use_db=False)
+                except:
```
gbrammer, line 3190: make this `except TypeError:` to explicitly catch "got an unexpected keyword argument"

grizli/prep.py
```diff
@@ -3179,11 +3219,19 @@ def process_direct_grism_visit(direct={},
...
+                img = jwst_utils.img_with_wcs(file)
+                img.save(file)
+                temp_hdu = pyfits.open(file,mode='update')
+                temp_hdu[0].header['FILTER'] = temp_hdu[0].header['FILTER_INFO']
+                temp_hdu[0].header['EXP_TYPE'] = temp_hdu[0].header['EXPTYPE_INFO']
+                temp_hdu.flush()
+            else:
+                try:
+                    updatewcs.updatewcs(file, verbose=False, use_db=False)
+                except:
```
gbrammer, line 3232: Same as above.  Not great to have things duplicated from the direct image processing, but the handling does need to be different here.  Also will have to update here if you change the placeholder header keywords like I mentioned above.

grizli/prep.py
gbrammer: Or another cleaner way, here specified as a YAML string but that could also be in a file like `grizli/data/jwst_references.yml`.  But then this starts to look like what crds is doing anyway....

```
import yaml # at the top

_JWST_FLATS = """
NIRISS:
  GR150R:
    F115W: jwst_niriss_flat_0202.fits
    F150W: jwst_niriss_flat_0187.fits
  GR150C:
    F115W: xxx.fits
NIRCAM:
  F115W:
      GRISMR: yyy.fits
      GRISMC: zzz.fits
"""
jwst_flats = yaml.safe_load(_JWST_FLATS)
flat_file = jwst_flats[INSTRUMENT][FILTER][PUPIL]

File: grizli/prep.py [hunk 54]
```diff
@@ -3222,6 +3270,94 @@ def process_direct_grism_visit(direct={},
...
+            hdu[1].header['NGOODPIX'] = -99
+            hdu[1].header['EXPNAME'] = hdu[0].header['EXPOSURE']
+            hdu[1].header['MEANDARK'] = -99
+            hdu[1].header['IDCSCALE'] = 0.065 ## get this automatically when pixelscale is in image headers
+            hdu[0].header['PHOTFLAM'] = phot_keywords[hdu[0].header['PUPIL']][0]
+            hdu[0].header['PHOTFNU'] = phot_keywords[hdu[0].header['PUPIL']][1]
+            hdu[0].header['PHOTPLAM'] = phot_keywords[hdu[0].header['PUPIL']][2] * 10000 # microns to angstroms
+            gain_file = steps_det1.gain_scale.get_reference_file(file, 'gain')
+            gain_im = pyfits.open(gain_file)
+            if (hdu[0].header['FILTER'] == 'GR150C') & (hdu[0].header['PUPIL'] == 'F115W'):
```
gbrammer, line 3335: See a comment above for specifying the flat list as a dictionary, which will need to be specified in just one place for use here and also where it came up above.  Maybe a more elegant solution would be having an explicit separate function that returns all necessary keywords based on an input header, like 

```python
_PHOTKEYS_YML = """
PHOTFLAM: 
  NIRISS:
    F090W: 1.098934e-20
    F115W: 1.23456
  NIRCAM:
    F090W: 3.4567

PHOTFNU:
  NIRISS:
    F090W: 9.99999
    F115W: 9.99999
  NIRCAM:
    F090W: 9.999
        
PHOTPLAM:
  NIRISS:
    F090W: 9.99999
    F115W: 9.99999
  NIRCAM:
    F090W: 9.999
"""
_PHOTKEYS = yaml.safe_load(_PHOTKEYS_YML)

def get_hst_photom_keywords(hdu):
    get PHOTKEYS based on hdu[0].header
```

And the HDU updates could be done in separate functions that would be easier to maintain and extend, e.g., in pseudocode

```python
def jwst_hdu_like_hst(hdu):
    if niriss:
        niriss_hdu_like_hst(hdu)
    elif nircam:
        nircam_hdu_like_hst(hdu)
    elif miri:
        miri_hdu_like_hst(hdu)

def niriss_hdu_like_hst(hdu):
    # ... set keywords etc.
   
...
```
And if the gain correction needs to be applied,  it should be done above where you do the flat correction.  And better still, the flat and gain corrections should be done by the `jwst` pipeline steps themselves.

grizli/prep.py
gbrammer: I put some functions to compute the rotations in [grismconf.py](https://github.com/gbrammer/grizli/blob/master/grizli/grismconf.py#L590)

grizli/prep.py
```diff
@@ -5595,3 +5797,26 @@ def extract_fits_log(file='idk106ckq_flt.fits', get_dq=True):
...
+    Returns
+    -------
+    isJWST : bool
+    '''
+
+    hdu = pyfits.open(infile)
+    try:
+        isJWST = (hdu[0].header['TELESCOP'] == 'JWST')
+        hdu.close()
+    except:
```
gbrammer, line 5819: what exception is raised here?  If `TELESCOP` not found?  If so, make the test explicit:
```python
with pyfits.open(infile) as hdu:
    if 'TELESCOP' in hdu[0].header:
        isJWST = (hdu[0].header['TELESCOP'] == 'JWST')
    else:
        isJWST = False
```

And it also occurs to me that `isJWST` throughout should probably just be handled rather by `INSTRUMENT`, since, for example, you have a different array shape for MIRI.

So rather, you just pull out `INSTRUMENT` and then various tests would be 
changed from `if isJWST` to `if _instrument in ['NIRISS','NIRCAM']`

grizli/prep.py
```diff
@@ -145,6 +155,30 @@ def fresh_flt_file(file, preserve_dq=False, path='../RAW/', verbose=True, extra_
                                calib_types=calib_types,
                                verbose=False)
 
+    if filter in ['GR150C', 'GR150R']:
```
gbrammer, line 158: Getting these directly from CDBS can wait, but this could be cleaned up into a dictionary for now,  like 
```python
niriss_flats = {'GR150R': 
                         {'F115W': 'jwst_niriss_flat_0202.fits', 
                          'F150W': 'jwst_niriss_flat_0187.fits', 
                          ...}, 
                'GR150C': ...
                }

flat_file = niriss_flats[FILTER][PUPIL]
```

Note that `jref` is usually for HST ACS (`jref` for WFC3), but OK to use as shorthand for now until it's fully implemented with CRDS.

Or another cleaner way, here specified as a YAML string but that could also be in a file like `grizli/data/jwst_references.yml`.  But then this starts to look like what crds is doing anyway....

```python
import yaml # at the top

_JWST_FLATS = """
NIRISS:
  GR150R:
    F115W: jwst_niriss_flat_0202.fits
    F150W: jwst_niriss_flat_0187.fits
  GR150C:
    F115W: xxx.fits
NIRCAM:
  F115W:
      GRISMR: yyy.fits
      GRISMC: zzz.fits
"""
jwst_flats = yaml.safe_load(_JWST_FLATS)
flat_file = jwst_flats[INSTRUMENT][FILTER][PUPIL]
```

grizli/prep.py
gbrammer, line 9: What's this?

grizli/pipeline/photoz.py
gbrammer, line 274: Eventually add all other filters...

grizli/pipeline/default_params.py
gbrammer, line 17: Please wrap at <79 characters here and below for PEP8, perhaps wrapping at logical breaks like between F200W and F277W for NIRISS (F277W and redder are in the `FILTER` wheel while the bluer filters are in the `PUPIL` wheel....) and between the `W`, `M` and `N` filters for NIRCam.

grizli/multifit.py
```diff
@@ -4003,7 +4008,7 @@ def drizzle_grisms_and_PAs(self, size=10, fcontam=0, flambda=False, scale=1, pix
                                       fcontam=0, fill_wht=True, ds9=None,
                                       mask_segmentation=mask_segmentation)
 
-            kern = h_kern[1].data[:, h['CRPIX1']-1-size:h['CRPIX1']-1+size]
+            kern = h_kern[1].data[:, int(h['CRPIX1'])-1-size:int(h['CRPIX1'])-1+size]
```
gbrammer, line 4011: I guess I was assuming that `CRPIX` was always integers, which isn't true (0.0 is the center of the pixel in the FITS standard, ±0.5 are the edges).  This is probably fine, but would be worth checking if it matters here (I think it could be `1024.5` for JWST?).  By that I mean that you're right that the slice needs to be integers, but check if the slice is offset by a pixel  one way or the other from expected.

grizli/multifit.py
gbrammer, line 3719: There's something inconsistent here with mixing JWST dispersers and bandpass filters.  Maybe we should call the dispersive element `dispname` throughout, which will be `FILTER` for HST, `PUPIL-FILTER` for NIRISS and `FILTER-PUPIL` for NIRCam, with a function in `utils` that generates this from a passed `Header` or `ImageHDU`.

grizli/model.py
```diff
@@ -3875,18 +3879,18 @@ def __init__(self, flt=None, beam=None, conf=None,
         # self.min_sens = min_sens
         # self.mask_resid = mask_resid
 
-        self._parse_from_data(**self._parse_params)
+        self._parse_from_data(isJWST=isJWST, **self._parse_params)
 
     def _parse_from_data(self, contam_sn_mask=[10, 3], min_mask=0.01,
-                         seg_ids=None, min_sens=0.08, mask_resid=True):
+                         seg_ids=None, min_sens=0.08, mask_resid=True, isJWST=False):
```
gbrammer, line 3885: Are these `isJWST` still needed?

grizli/jwst_utils.py
gbrammer, line 132: FITS header keywords generally need to be 8 characters or less, though I'm not too careful with that elsewhere.  Maybe `OFILTER` and `OEXPTYPE`? (those are "Oh's", not zeros)

grizli/multifit.py
```diff
@@ -1411,7 +1412,8 @@ def __init__(self, beams, group_name=None, fcontam=0., psf=False, polyx=[0.3, 2.
 
         if isinstance(beams, str):
             self.load_master_fits(beams, verbose=verbose)
-
+            # check if isJWST
+            isJWST = prep.check_isJWST(beams)
```
gbrammer, line 1416: Also appears in the loop just below

grizli/prep.py
gbrammer, line 3190: make this `except TypeError:` to explicitly catch "got an unexpected keyword argument"

grizli/prep.py
gbrammer, line 3232: Same as above for `except TypeError`.  Not great to have things duplicated from the direct image processing, but the handling does need to be different here.  Also will have to update here if you change the placeholder header keywords like I mentioned above.

grizli/prep.py
```diff
@@ -4451,19 +4639,26 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
     data_fixed = []
     for file in bg_fixed:
         im = pyfits.open('{0}/CONF/{1}'.format(GRIZLI_PATH, file))
-        sh = im[0].data.shape
-        data = im[0].data.flatten()/flat
+        if isJWST:
```
gbrammer, line 4642: the only thing different here is the HDU extension, so make 
```python
if isJWST:
    sci_ext = 1
else:
    sci_ext = 0

sh = im[sci_ext].data.shape
data = im[sci_ext].dat.flatten() / flat
```

grizli/prep.py
```diff
@@ -4612,7 +4809,10 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
     fig = plt.figure(figsize=[6., 6.])
     ax = fig.add_subplot(111)
 
-    im_shape = (1014, 1014)
+    if isJWST:
```
gbrammer, line 4812: See above, i.e., MIRI will be different, though I guess we (probably) won't be doing grism background for MIRI
```python
... 
if _instrument in ['NIRISS', 'NIRCAM']:
    ...
```

grizli/prep.py
```diff
@@ -4716,15 +4916,16 @@ def visit_grism_sky(grism={}, apply=True, column_average=True, verbose=True, ext
...
+        fp = open(file.replace('_flt.fits', '_column.dat').replace('_rate.fits', '_column.dat'), 'wb')
         fp.write(b'# column obs_resid ok resid uncertainty\n')
         np.savetxt(fp, np.array([xmsk, yres, yok*1, y_pred-bg_sky, gp_sigma]).T, fmt='%.5f')
         fp.close()
 
         if apply:
             # Subtract the column average in 2D & log header keywords
-            gp_res = np.dot(y_pred[:, None]-bg_sky, np.ones((1014, 1)).T).T
             flt = pyfits.open(file, mode='update')
+            imshape = flt[1].data.shape[0]
```
gbrammer, line 4927: Are we sure that the JWST grism exposures are being rotated correctly in this `prep` processing so that the column subtraction is truly columns parallel to the dispersion axis as expected?

I put some functions to compute the rotations in [grismconf.py](https://github.com/gbrammer/grizli/blob/master/grizli/grismconf.py#L590)

grizli/prep.py
gbrammer, line 5819: What exception is raised here?  If `TELESCOP` not found?  If so, make the test explicit:
```python
with pyfits.open(infile) as hdu:
    if 'TELESCOP' in hdu[0].header:
        isJWST = (hdu[0].header['TELESCOP'] == 'JWST')
    else:
        isJWST = False
```

And it also occurs to me that `isJWST` throughout should probably just be handled rather by `INSTRUMENT`, since, for example, you have a different array shape for MIRI.

So rather, you just pull out `INSTRUMENT` and then various tests would be 
changed from `if isJWST` to `if _instrument in ['NIRISS','NIRCAM']`

grizli/tests/test_jwst.py
gbrammer, line 20: This will fail in the online tests. You could make a test to quit gracefully if `not os.path.exists(beams_file)`, or if the file isn't too big you could put it in `grizli/tests/data` in the repo. If you do the latter, it would be good to call it something like niriss_jw01....beams.fits.

grizli/pipeline/auto_script.py
```diff
@@ -442,7 +442,7 @@ def go(root='j010311+131615',
 
     visits, all_groups, info = parsed
     run_has_grism = utils.column_string_operation(info['FILTER'],
```
gbrammer, line 444: I didn't know about `np.in1d`, which is much cleaner. So

```python
run_has_grism = np.in1d(info['FILTER'], ['G141', 'G102', 'G800L', 'GR150C', 'GR150R']).sum() > 0
# is PUPIL in info?
run_has_grism |= np.in1d(info['PUPIL'], ['GRISMR', 'GRISMC']).sum() > 0
```

grizli/pipeline/auto_script.py
```diff
@@ -1389,20 +1398,21 @@ def parse_visits(field_root='', RAW_PATH='../RAW', use_visit=True, combine_same_
...
                                   use_visit=use_visit, max_dt=max_dt, 
-                                  visit_split_shift=visit_split_shift)
+                                  visit_split_shift=visit_split_shift,
+                                  isJWST=isJWST)
+    
 
     # Don't run combine_minexp if have grism exposures
-    grisms = ['G141', 'G102', 'G800L', 'G280']
+    grisms = ['G141', 'G102', 'G800L', 'G280', 'GR150C', 'GR150R']
     has_grism = utils.column_string_operation(info['FILTER'], grisms,
```
gbrammer, line 1416: See above for `run_has_grism`

grizli/pipeline/auto_script.py
```diff
@@ -1892,6 +1904,18 @@ def preprocess(field_root='j142724+334246',  HOME_PATH='/Volumes/Pegasus/Grizli/
     # Clean up
     if clean:
         clean_prep(field_root=field_root)
+    # Rename instrument back to JWST stuff
+    # add an if statement later for this
+    if isJWST:
+        for file in direct['files']:
+            hdu = pyfits.open(file, mode='update')
+            hdu[0].header['INSTRUME'] = 'NIRISS'
```
gbrammer, line 1912: Save the original keyword above when you change it (e.g., `OINSTRUM`) and change it back to that, e.g., NIRCam / MIRI.

grizli/pipeline/auto_script.py
```diff
@@ -2429,13 +2453,25 @@ def load_GroupFLT(field_root='j142724+334246', PREP_PATH='../Prep', force_ref=No
...
+    masks = {'g141' : [info['FILTER'] == 'G141', 'G141'],
+    'g102' : [info['FILTER'] == 'G102', 'G102'],
+    'g800l' : [info['FILTER'] == 'G800L', 'G800L'],
+    'gr150cf115w' : [(info['PUPIL'] == 'F115W') & (info['FILTER'] == 'GR150C'), 'GR150C', 'F115W'],
+    'gr150cf150w' : [(info['PUPIL'] == 'F150W') & (info['FILTER'] == 'GR150C'), 'GR150C', 'F150W'],
+    'gr150cf200w' : [(info['PUPIL'] == 'F200W') & (info['FILTER'] == 'GR150C'), 'GR150C', 'F200W'],
+    'gr150rf115w' : [(info['PUPIL'] == 'F115W') & (info['FILTER'] == 'GR150R'), 'GR150R', 'F115W'],
+    'gr150rf150w' : [(info['PUPIL'] == 'F150W') & (info['FILTER'] == 'GR150R'), 'GR150R', 'F150W'],
+    'gr150rf200w' : [(info['PUPIL'] == 'F200W') & (info['FILTER'] == 'GR150R'), 'GR150R', 'F200W']}
+
```
gbrammer, line 2473:
```python
masks = {}
for gr in ['G141', 'G102', 'G800L']:
    masks[gr.lower()] = (info['FILTER'] == gr, gr, '')

for gr in ['GR150R', 'GR150C']:
    for filt in ['F090W, ''F115W', 'F150W', 'F200W']:
        key = f'{gr.lower()}-{filt.lower()}'
        masks[key] = (info['PUPIL'] == filt) & (info['FILTER'] = gr), 
                      gr, filt)

# NIRCAM ....
```

grizli/pipeline/auto_script.py
```diff
@@ -2889,7 +2846,7 @@ def extract(field_root='j142724+334246', maglim=[13, 24], prior=None, MW_EBV=0.0
 
         #mb = multifit.MultiBeam(beams, fcontam=fcontam, group_name=target, psf=False, MW_EBV=MW_EBV, min_sens=min_sens)
 
-        mb = multifit.MultiBeam(beams, fcontam=fcontam, group_name=target, psf=False, MW_EBV=MW_EBV, sys_err=sys_err, min_mask=min_mask, min_sens=min_sens)
+        mb = multifit.MultiBeam(beams, fcontam=fcontam, group_name=target, psf=False, isJWST=isJWST, MW_EBV=MW_EBV, sys_err=sys_err, min_mask=min_mask, min_sens=min_sens)
```
gbrammer, line 2849: Is this still needed?

grizli/pipeline/auto_script.py
```diff
@@ -4006,7 +3976,7 @@ def make_combined_mosaics(root, fix_stars=False, mask_spikes=False, skip_single_
         if fill_mosaics == 'grism':
             # Only fill mosaics if grism filters exist
             has_grism = utils.column_string_operation(info['FILTER'],
```
gbrammer, line 3978: See again comment above for `has_grism`.  Maybe just set `info.meta['HAS_GRISM']` once when it's generated to avoid having to test it every time

grizli/utils.py
```diff
@@ -1371,14 +1384,16 @@ def calc_header_zeropoint(im, ext=0):
...
-    if 'PHOTFNU' in header:
-        ZP = -2.5*np.log10(header['PHOTFNU'])+8.90
-        ZP += 2.5*np.log10(scale_exptime)
-    elif 'PHOTFLAM' in header:
+    if 'PHOTFLAM' in header:
         ZP = (-2.5*np.log10(header['PHOTFLAM']) - 21.10 -
               5*np.log10(header['PHOTPLAM']) + 18.6921)
 
         ZP += 2.5*np.log10(scale_exptime)
+        print(np.log10(header['PHOTFLAM']), np.log10(header['PHOTPLAM']), scale_exptime)
```
gbrammer, line 1392: Should this be logged or is it just debugging?

gbrammer (commented):
A number of suggestions throughout

grizli/utils.py
victoriastrait: Just debugging. Removed here: b531805f5953180248c849306194ebdda2837e21

grizli/pipeline/auto_script.py
victoriastrait: Yes, to avoid a circular import

grizli/pipeline/auto_script.py
victoriastrait: Done here: a8db578018fc195779267b6462d6006b0837baab

grizli/jwst_utils.py
victoriastrait: fixed (also in prep where I change them back): b663fc80f900f3edeab7a900b18fcf8d24b07e2e

grizli/multifit.py
victoriastrait: It has to check differently for the second case (it checks beams if beams is a string and beams[0] if beams is a list of strings)

grizli/pipeline/default_params.py
victoriastrait: Done here: 927d4cfff34c7cf1e64f2b554060f9de7286ce5a

grizli/prep.py
victoriastrait: Done here: 56af2ae77b750072824b7a7dcf92c47b43d3d4bd

grizli/prep.py
victoriastrait: Done here: 22b1a7847907cfedec10eb55d3566a21fb9e3b4d

grizli/prep.py
victoriastrait: Done in the same commit as above: 22b1a7847907cfedec10eb55d3566a21fb9e3b4d

grizli/pipeline/auto_script.py
victoriastrait: Done here: 28b92ae72465cf463a9a89570daa152f57a0fe16

grizli/pipeline/auto_script.py
victoriastrait: Done here: ef053b3aed8f8e694f9e8488e76919c9f68b04ca

grizli/tests/test_jwst.py
victoriastrait: Fixed here: ba9d290ea7488b9bfadfd384e6bf6df03e248054 and here: dcd4ca6daf7f9c9ae4eec38ce3a05321f01fb24a

grizli/tests/test_jwst.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+import glob
+import unittest
+
+import numpy as np
+from .. import utils, prep, multifit, fitting, GRIZLI_PATH
+
+
+class JWSTFittingTools(unittest.TestCase):
+
```
gbrammer, line 13: The test action is failing becauase it can't find the NIRISS configuration file(s).  This class should have a first method that checks for the necessary config files and fetch them if not, analogous to https://github.com/gbrammer/grizli/blob/master/grizli/tests/test_fits.py#L17.

This could be another option in [``utils.py``](https://github.com/gbrammer/grizli/blob/master/grizli/utils.py#L5729)  (`get jwst`) to add [jwst-grism-conf.tar.gz](https://github.com/gbrammer/grizli-config/blob/main/jwst-grism-conf.tar.gz) to the list of tarfiles to download and extract.

gbrammer (commented):
After fixing the test problem, if you can be ready to file an Issue requesting cleaning up to parse the calib files with the pipeline rather than the hard-coded `if/else` sequences, I think we can merge this in as soon as the test action passes.

grizli/tests/test_jwst.py
victoriastrait: I've tried to do this here: 8ef42e7f9e5469244df1ca98ce56384a634b1641 Let me know if it works

grizli/utils.py
```diff
@@ -5751,6 +5765,9 @@ def fetch_config_files(ACS=False, get_sky=True, get_stars=True, get_epsf=True):
 
     tarfiles += [f'{BASEURL}/ACS.WFC.CHIP1.Stars.conf', 
                  f'{BASEURL}/ACS.WFC.CHIP2.Stars.conf']
+
+    if get_jwst:
+        tarfiles += [f'{BASEURL}/jwst-grism-conf.tar.gz']
```
gbrammer, line 5770: Good!  Could you do one more thing and make fetching wfc3 not implicit?  That is, add a `fetch_wfc3` option with default `True` and append the `V4.32` files to `tarfiles` if it's set.  And then also the `ACS.WFC.CHIP?.Stars.conf` rows should be within the `ACS` block.  And finally I guess the option `ACS` should be `get_acs` and you can provide some back compatibility with 

```python
def fetch_config_files(..., **kwargs):
    ...
    if 'ACS' in kwargs:
        fetch_acs = kwargs['ACS']
    ...
```

gbrammer (commented):
The test still fails because the assert line checking if the file exists fails.

grizli/utils.py
gbrammer: NB, the [installation docs](https://github.com/gbrammer/grizli/blob/master/docs/grizli/install.rst) would need to be updated with the new `get_acs` and also `get_jwst` keywords.

grizli/prep.py
victoriastrait: I've added a check for rot90: d42e1240cfbd85a60beb56e8237db7d32c19938c

In the cases we've had so far, it has seemed to be ok. However, I'm not totally sure I switched the axes correctly for the case where we'd need to do that.

gbrammer: I don't see it in the diff here, but you'd need to change the axes along with the statistics are averaged [here](https://github.com/gbrammer/grizli/blob/master/grizli/prep.py#L4626).

grizli/utils.py
victoriastrait: Fixed. Is the test working now? 81e00bf075d9ff4250ebd77ab5df0ae2ba41629b

gbrammer:
My new test on the visit parsing breaks because your function now requires `info`.  Can you please add it [here](https://github.com/gbrammer/grizli/blob/master/grizli/tests/test_autoscript.py#L40)?

gbrammer:
The code is working on basic NIRISS test data without failures and the WFC3 demo processes fully as before, so we'll merge this in and keep working on individual fixes and modifications with individual Issues / PRs.

Thanks for all of the great work getting this through.  This is a huge improvement and gets us in a much better position to receive the data that will be coming in a matter of weeks!

PR #78, opened 2022-05-25
Fix some issues in JWST headers and parsing to improve AstroDrizzle compatibility

gbrammer:
AstroDrizzle requires some more detailed aspects in the FITS files in order to "pretend" that the JWST exposures are HST WFC3/IR.  For example, they require a TIME HDU extension that describes the exposure time, per pixel.  This can be an empty HDU with a single value indicated in the `PIXVALUE` keyword.

### DIFF
--- a/grizli/jwst_utils.py
+++ b/grizli/jwst_utils.py
@@ -106,6 +106,8 @@ def img_with_flat(input, verbose=True, overwrite=True):
     """
     Apply flat-field correction if nessary
     """
+    import gc
+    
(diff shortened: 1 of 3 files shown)


gbrammer:
@victoriastrait these fixes should allow you to get to the end of the "GroupFLT" contamination modeling in the pipeline demo, which includes making the visit-level mosaic in the direct imaging filter and deriving a catalog from it.

The grism model doesn't look great because of the residual gradient across the simulations, but the spectra are mostly in the right place.

Issue #87, opened 2022-06-08
Need to distinguish NIRISS and NIRCam/SW filters with the same name


gbrammer:
Otherwise they'll be processed together, e.g., when making the final drizzle combined images across filters.

For filters in common between ACS/WFC and WFC3/UVIS, a `u` is appended to the filter names, but this is a bit ugly.  

[utils.get_hst_filter](https://github.com/gbrammer/grizli/blob/master/grizli/utils.py#L1211) would be the place to do this, but elsewhere the functions that parse the filter name from the visit product name assume that the filter name ends in a single character.

gbrammer:
This isn't quite the same issue, but the parse_visits functions should also distinguish between the different detectors

PR #89, opened 2022-06-14
Rename get_hst_filter to parse_filter_from_header, include NIRCam detectors in visit parsing

gbrammer:
Rename the function `utils.get_hst_filter` to `utils.parse_filter_from_header` throughout, since it parses headers from other telescopes.  

That new function includes a keyword argument `jwst_detector=False` that prepends the JWST detector to the filter string, which is useful mostly for making unique visit product names in `utils.get_flt_info` and `utils.parse_flt_files`.  

**Note:** The detector name is extracted from the ``DETECTOR`` keyword, but in NIRCam the LW detector name is changed from "LONG" to "5", e.g., "nrca5", to better match the FITS filenames.

So, product names are something like ``unknown-02561-001-036.0-nrcb1-f115w-clear`` and ``unknown-02561-001-036.0-nrcb5-f277w-clear``.

This PR also fixes a bug where the JWST header keywords were being set extra times before passing to drizzle.

### DIFF
--- a/grizli/aws/aws_drizzler.py
+++ b/grizli/aws/aws_drizzler.py
@@ -918,7 +918,7 @@ def drizzle_images(label='macs0647-jd1', ra=101.9822125, dec=70.24326667, pixsca
 
                 filt_i = file.split('-')[-1].split('_dr')[0]
             else:
-                filt_i = utils.get_hst_filter(hdu_i[0].header)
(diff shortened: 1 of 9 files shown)


gbrammer:
This addresses part of the visit naming issue at https://github.com/gbrammer/grizli/issues/87.

PR #104, opened 2022-07-25
Update niriss config

gbrammer:
Use new config files with better positioning of the NIRISS 0th order spectra


gbrammer:
@victoriastrait I pushed some new config files to https://github.com/gbrammer/grizli-config/blob/main/niriss.conf.220725.tar.gz and this update uses them.

PR #105, opened 2022-08-12
Process JWST with rate files

gbrammer:
Do processing with JWST rate files where you can explicitly specify a `CRDS_CONTEXT` to use the latest `flat` and `photom` calibrations.

The default `CRDS_CONTEXT` version is set in `grizli.jwst_utils.CRDS_CONTEXT` and this is used when initializing the files in the copy from `RAW` > `Prep` with `prep.fresh_flt_files`.

The `flat` and `photom` files that were used are stored in the header in `R_FLAT` and `R_PHOTOM` keywords, respectively.

### DIFF
--- a/grizli/aws/visit_processor.py
+++ b/grizli/aws/visit_processor.py
@@ -1115,7 +1115,8 @@ def process_visit(assoc, clean=True, sync=True, max_dt=4, combine_same_pa=False,
     os.environ['orig_jref'] = os.environ.get('jref')
     set_private_iref(assoc)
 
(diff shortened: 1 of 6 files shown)


gbrammer:
MIRI updates:
==========
- don't mask the full left side of the detector
- make a sky flat from visit exposures (works best for many exposures and large dithers)
- the sky flat incorrectly mixes issues with the MIRI darks and flats, but gives aesthetically improved results

Others
=====
- Don't update tables when `visit_processor.process_visit(sync=False)`
- Remove inverse SIP coefficients from headers

PR #107, opened 2022-08-26
Add JWST photometry correction

gbrammer:
As of late August 2022, a number of groups have found that there still seem to be large corrections to the individual NIRCam SCA photometric calibration required, even after the CRDS updates in late July.  There is a series of ongoing calibration programs that will establish the absolute calibration of all of the modes at high precision.  For now, however, those calibration data are incomplete.  

I have derived some corrections to the `jwst_0942.pmap` context using the following:

1) The flux calibration standard `J1743045` suggests that the absolute NIRISS calibration is quite good, with just a correction of 1/1.04 required for F200W
2) The flux standard also suggests that NIRCam LW broadbands are fairly well calibrated, perhaps with the exception of a 1/0.912 correction needed for F444W
3) The flux standard data are currently only available for the B module in some filters.  To calibrate the A module and also the additional broadband SW filters, I derived *relative* corrections using the wealth of data taken of the LMC astrometric calibration field, where certain patches of the field have been observed by more than one NIRCam SCA.  The corrections are derived by measuring photometry in the NIRISS broad bands and then calculating scale factors of all of the NIRCam SW SCAs relative to these.
4) Relative A/B corrections for the NIRCam LW broad-band filters are also derived from the LMC data.

The tabulated values in `grizli/data/photom_correction.yml` are the median ratios of the derived flux densities in the indicated filters to some reference, either relative or absolute.  To apply the corrections relative to exposures calibrated with the current CRDS_CONTEXT, the values should be *divided* from the SCI and ERR arrays, which is now implemented in `utils.drizzle_from_visit`.

The LMC mosaics are available at https://s3.amazonaws.com/grizli-v2/JwstMosaics/lmc-astrometric-field/lmc.html.  The full notebook used to derive the LMC corrections is at https://s3.amazonaws.com/grizli-v2/JwstMosaics/lmc-astrometric-field/astrometric-field-photometry.ipynb.

The figures below show the distribution of flux ratios in the NIRCam LW and SW filters, where the median values are stored in the correction table.

Module ratios for **NIRCam LW**
nircam_LW_astrometric_field

**NIRCam SW ratios relative to NIRISS**

nircam_to_niriss-F090W_astrometric_field
nircam_to_niriss-F115W_astrometric_field
nircam_to_niriss-F150W_astrometric_field
nircam_to_niriss-F200W_astrometric_field

The *multiplicative* corrections are summarized below:
```
# det filt  pupil  mjsr_0942    gbr    ila  mjsr_gbr
[...]

PR #108, opened 2022-09-06
Update F410M correction based on PRIMER overlap

gbrammer:
The v4.0 catalogs include the photometry corrections described in https://github.com/gbrammer/grizli/pull/107.  That update included an interpolated value for the F410M filter, which wasn't sampled in the test data.  The eazy zeropoint maps in those catalogs suggest that that interpolated value isn't quite correct, as in the attached figure.

There is a slight overlap between the A/B modules in the F410M PRIMER mosaic that suggests A/B ~ 1.01 rather than the somewhat larger corrections in the wide filters.  Absolute correction remains TBD.

download-4


gbrammer:
The new eazy spatial diagnostic image no longer shows a strong imprint of the two LW channels
download-7

Issue #113, opened 2022-10-13
'PermissionError: [Errno 1] Operation not permitted' when using autoscript.go()


njcleri:
When running the following code to extract spectra for G141 observations of EGS, I get the following error: 
`PermissionError: [Errno 1] Operation not permitted: '/tmp/aegis-01-bhj-39-000.0-f140w_tmpwcs.fits'`
Any idea where this could be coming from? Is the query not getting files correctly?

```
import glob
import time
import os
import sys
import yaml

import numpy as np
import matplotlib.pyplot as plt
from IPython.display import Image

from astropy.io import fits
from astropy.table import Table

import drizzlepac

import warnings
warnings.filterwarnings('ignore')

from mastquery import query, overlaps

import grizli
from grizli.pipeline import auto_script
from grizli import utils, fitting, multifit, prep
from grizli.pipeline.auto_script import get_yml_parameters
import grizli.utils
# HST calibs to iref/iref
# set get_acs=True below to get files necessary for G800L processing
grizli.utils.fetch_default_calibs()
# config files to $GRIZLI/CONF
# set get_jwst=True to get config files for jwst processing
grizli.utils.fetch_config_files()
grizli.utils.symlink_templates(force=False)
HOME_PATH = os.getcwd()
print(HOME_PATH)

parent = query.run_query(box=[214.6500, 52.6500, 50], proposal_id=[12177, 12547, 13063, 12063], 
                         instruments=['WFC3/IR'], 
                         filters=['F140W','G141'])
                         
tabs = overlaps.find_overlaps(parent, buffer_arcmin=10, proposal_id=[12177, 12547, 13063, 12063], instruments=['WFC3/IR'], filters=['F140W', 				      'G141'], close=False)

print(glob.glob('*footprint.fits'))
foot_files = np.array(glob.glob('*footprint.fits'))
print('Footprint files: ', foot_files)

print('\n# id            ra         dec        e(b-v)   filters')
for tab in tabs:
    print('{0}  {1:.5f}  {2:.5f}   {3:.4f}   {4}'.format(tab.meta['NAME'], tab.meta['RA'], 
                                                 tab.meta['DEC'], tab.meta['MW_EBV'],
                                                  ','.join(np.unique(tab['filter']))))

root_list = [tab.meta['NAME'] for tab in tabs]
print(root_list)

## Read the default parameters that can be edited and passed en-masse to `auto_script.go`
kwargs = get_yml_parameters()
print(list(kwargs.keys()))

[...]

PR #115, opened 2022-10-19
Remove instances of writing files to /tmp

gbrammer:
Remove a few examples where temporary files were written to the `/tmp` directory.

The PR also implements closing out dangling `astropy.io.fits.open` objects that are no longer needed throughout.


gbrammer:
Removing the `/tmp` files is meant to address https://github.com/gbrammer/grizli/issues/113.

PR #118, opened 2022-11-14
Ahw changes

ivastar:
Hi Gabe,

(CC @jdavies-st, @BSGalvan)

This pull requests makes a number of changes that improve the packaging and installation of `grizli`. The changes include:

- Make changes such that `pip` installs all the dependencies together with the package. All dependencies are specified in `setup.cfg`. Optional dependencies (e.g., `test` and `aws`) are only installed when the flag is used. There is no longer a need to create an environment and manage dependencies with conda (environment is still recommended).
- The package is automatically built upon installation using modern PEP517-compliant packaging tools. 
- Remove custom versioning code based on git tag, use the standard `setuptools-scm` instead.  Works the same way, but maintained by the folks who run pypi.org.
- Replace `lacosmicx` with `astroscrappy`: same developer, same code, same API but maintained.
- Update the function that symlinks the templates to symlink the whole directory rather than individual files.
- Add a GitHub actions workflow which builds the wheels and (upon a tagged release) publishes to PyPI. You need to create and save PyPI token in order for this to work. See instructions [here](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/).
- Update the installation docs.
- Fix issues with tests.
- Minor clean-up.
 
One issue we couldn't resolve is the installation of `eazy-py`. Currently (in this PR) `eazy` is being pip-installed (via `setup.cfg`) but also the manual instructions are in the documentation. Seems like `eazy-py` is released to PyPI  but the templates from the original [eazy-photoz](https://github.com/gbrammer/eazy-photoz) repo are not downloaded with that installation. This can be fixed by (for example) adding a function that downloads the templates only (if this is all you need from the `eazy-photoz` repo) rather than making it a subpackage. It doesn't make sense to release `eazy-py` every time you update the templates and filters. Happy to edit this pull request (if you prefer to keep the manual `eazy-py` installation) or make a pull request to `eazy-py` with a proposed solution.

All manual installations (except for `hstcal`) can be removed if `tristars` and `reprocess_wfc3` are also released to PyPI and `pyregion` is replaced with `astropy.regions`. `pyregion` seems to be suffering maintenance issues and while it pip-installs fine on OSX it fails to install in the test CI. Happy to help fix all of these, the effort will be fairly minor (well, not sure about `pyregion`).

We did some minor tests. The tests in the repo run and we ran through some HST and JWST data and made sure at least that minimum functionality works. 

Happy to iterate on this. 

Cheers,
-Iva


jdavies-st:
Here are the main PRs that made it into this Astro Hack Week (AHW) branch.

https://github.com/ivastar/grizli/pull/6
https://github.com/ivastar/grizli/pull/7
https://github.com/ivastar/grizli/pull/12
https://github.com/ivastar/grizli/pull/20
https://github.com/ivastar/grizli/pull/21
https://github.com/ivastar/grizli/pull/22
https://github.com/ivastar/grizli/pull/23

More detail and discussion in these PRs, including CI results.

gbrammer:
Thanks a lot for all of this, @ivastar,  @jdavies-st, @BSGalvan!  This should make everything much easier both for us to maintain and for others to use.    One of my `actions` tests failed as you can see above, looks like related to the `version.py` file that is created with my `setup.py`.  That was always a bit of a hack for me to keep track of the git commit version, but the handling of `version.py` would be one key component to this modification.  

Can you give me an idea of the workflow for pushing new versions to pip, or will that now be done automatically for *every* commit to the `master` branch?

gbrammer:
And happy to switch to `astropy.regions`.  I expect that would be fairly simple, because `pyregion` is only used in a few places for generating pixel masks based on region files, assuming that functionality is provided by `astropy.regions`.

jdavies-st:
Hi Gabe, I didn't realize the JWST workflow also ran on CI.  It didn't on our end because it was only designed to work on pull requests against master.  Anyway, I've updated it to reflect the changes in the other one, and also removed redundant stuff like build and flake8 checks, which are done in the main CI workflow.  One could also merge these 2 workflows into a single workflow with two separate test jobs, one for non-JWST and one for JWST.  Up to you.

ivastar:
@jdavies-st fixed some issues with `pyregion` here: https://github.com/gbrammer/grizli/pull/118
If/once `pyregion` gets a new release with those fixes there will be no urgent need to switch to `astropy.regions` which seems to have a different API though still good to transition at some point soon. This PR works as it is and I can come back and do clean up once the `pyregion` is re-released to pypi.

gbrammer:
Thanks @jdavies-st and @ivastar .  The JWST CI test failed for missing `pyregion`, so would it be possible to put it back in the requirements for now?

jdavies-st:
I fixed `pyregion`.  Unfortunately it can't be included in `install_requires` because it fails installing on Linux currently.  `pip install pyregion` just doesn't work.  So it would mean `pip install grizli` would also fail on Linux if we included it.  Hopefully we'll have that fixed soon though, so the current steps in the jwst workflow are a stop-gap measure.

Should be fixed now.

jdavies-st:
> Can you give me an idea of the workflow for pushing new versions to pip, or will that now be done automatically for every commit to the master branch?

So this new publish-to-pypi workflow will run every time something is merged into `master`.  But that will just build the wheels.  This is not a bad idea, as it will let know if something has broken the building of the sdist or wheels.  Hopefully not!  But a nice sanity check.

It will also run if you manually make a release on Github.  The workflow for that is

1. Tag the commit you want to release.  Push it upstream.
2. Go to https://github.com/gbrammer/grizli/releases and click "Create a new release"
3. Choose the tag you just pushed to master.  You can also skip 1. and create a tag here too if you like.
4. Fill in Title and Description (optional) and hit "Publish release" button, which will trigger the workflow and delivery to PyPI so it can be pip installed.

ivastar:
To add to the previous comment by @jdavies-st, _technically_ you can trigger a publication to pypi with every merge to master. Your call. The way this is set up now, you (and any user) can still install it from github and get the latest changes but the pypi version is maybe released one a month and thus more stable?  The controls are in the [publish-to-pypi](https://github.com/gbrammer/grizli/blob/cb941ae46a7d3f587e5fb2aed81b43013818426f/.github/workflows/publish-to-pypi.yml) action. [This](https://github.com/gbrammer/grizli/blob/cb941ae46a7d3f587e5fb2aed81b43013818426f/.github/workflows/publish-to-pypi.yml#L6) controls when the action is run (currently pushes and pull requests to master are commented out) and [this line](https://github.com/gbrammer/grizli/blob/cb941ae46a7d3f587e5fb2aed81b43013818426f/.github/workflows/publish-to-pypi.yml#L54) controls the publication.

gbrammer:
Thanks for your continued work on this.  I just added [tristars v0.1](https://pypi.org/project/tristars/) and [reprocess_wfc3 v0.2.1](https://pypi.org/project/reprocess_wfc3) to PyPI, so could you please update ``setup.cfg`` and the installation docs accordingly?  Then I think it's ready to go.

jdavies-st:
Excellent!  Horray for `tristars` release.

PR #123, opened 2022-11-21
Change NIRCam wisp mask to use polygon

gbrammer:
Wisp regions defined in `data/nircam_wisp.yml`.


gbrammer:
This update also includes the utility function ``utils.pixel_polygon_mask`` to generate a pixel mask from a polygon array or ``polygon(...)`` region string.

PR #124, opened 2022-11-25
Add trace offsets for NIRCam F444W.

gbrammer:
- apply trace shifts derived from the first FRESCO visits
- add option to remove a `sep` background from the grism exposures

### DIFF
--- a/grizli/grismconf.py
+++ b/grizli/grismconf.py
@@ -996,7 +996,7 @@ class TransformGrismconf(object):
         
         # Read updated sensitivity files
         if 'V4/NIRCAM_F444W' in self.conf_file:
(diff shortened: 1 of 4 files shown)


gbrammer:
Here is a figure showing the derived trace offsets relative to the [GRISM_NIRCAM/V4](https://github.com/npirzkal/GRISM_NIRCAM/tree/master/V4) configuration files.  
f444w_trace_offsets

gbrammer:
And these are sensitivity curves that extend the `GRISM_NIRCAM/V4` 1st order sensitivity curves using the F444W filter throughput curve.
[extended_sensitivity_f444w.tgz](https://github.com/gbrammer/grizli/files/10092872/extended_sensitivity_f444w.tgz)

PR #127, opened 2022-12-13
Add pyregion dependency to install_requires

jdavies-st:
The `pyregion` package was released recently with fixes so it will install under modern python build systems.  (Thanks @larrybradley).  It now installs under Linux and can be added as a normal dependency to `grizli`.

Installation docs are updated appropriately.

@gbrammer if you decide to eventually take the time to switch over to [astropy/regions](https://github.com/astropy/regions), ping me in the PR, as your changes in `grizli` might be a nice starting point for docs changes in `pyregion` to show people how to do the switch over.  I'd be happy to do the docs changes over there.

### DIFF
--- a/.github/workflows/python-package-with-jwst.yml
+++ b/.github/workflows/python-package-with-jwst.yml
@@ -23,14 +23,12 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3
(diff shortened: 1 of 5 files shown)


gbrammer:
Excellent.  Thanks, @jdavies-st!

gbrammer:
Regarding the switch to `astropy.regions`, is there already functionality there for making masks based on headers / WCS?  I looked for a bit the other day when we were discussing and don't remember seeing anything.

PR #129, opened 2022-12-16
Convert code to use regions instead of pyregion

larrybradley:
This PR converts the code to use the `regions` package instead of `pyregion`.  While doing so, I also fixed a couple of indexing bugs (ds9 regions are always 1-indexed).

Followup to #127.

CC: @jdavies-st


gbrammer:
Thanks a lot, @larrybradley ! Glad to see it was a pretty straightforward swap.

PR #131, opened 2023-01-03
Fixing the installation docs

ivastar:
Two changes:
- fixing a formatting issue that I introduced in an earlier pull request
- updating the `eazy` installation instructions including the `eazy.fetch_eazy_photoz()` function


gbrammer:
Thanks, @ivastar !

PR #139, opened 2023-02-17
Fix bugs with rotated line maps

gbrammer:


gbrammer:
- Adds a N-E compass to line maps
- add offset to traces for line maps

PR #145, opened 2023-03-03
Update utils.py

JacobKosowski:
os.remove() only removes a selected file, though the code expects the removal of a directory. os.rmdir() is not a suitable substitute since /templates could contain files.

### DIFF
--- a/grizli/utils.py
+++ b/grizli/utils.py
@@ -2,6 +2,7 @@
 Dumping ground for general utilities
 """
 import os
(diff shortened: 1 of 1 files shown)


gbrammer:
Thanks!  Could I ask you a favor to send the same fix to [eazy-py](https://github.com/gbrammer/eazy-py)?

JacobKosowski:
Done!

PR #146, opened 2023-03-06
Functions for subtracting column averages along rotated axes for NIRCam star spikes

gbrammer:
Some NIRCam visits appear to show diffraction spikes from *very* bright stars that are as much as a few degrees away from the target field (e.g., the star Mira in PRIMER UDS).  This PR includes functions to compute column averages of combined images rotated with respect to the ``PA_APER`` instrument position angle.  The main off-axis diffraction spikes for NIRCam are at PA_APER±30 deg.

To turn this on, run the processing pipeline with, e.g., 

```python
>>> visit_processor.process_visit(assoc,
                                  prep_args={'angle_background_kwargs':
                                                  {'instruments': ['NIRCAM'],
                                                   'niter':3, 'threshold':1.8,
                                                   }
                                  })
```
This PR also implements looking for a ``BKG`` extension in the exposure files when making a mosaic, so that any derived background can be kept separate from the ``SCI`` arrays.

### DIFF
--- a/grizli/aws/lambda_handler.py
+++ b/grizli/aws/lambda_handler.py
@@ -814,13 +814,17 @@ def run_grizli_fit(event):
             grizli_db.add_redshift_fit_row(rowfile, table=dbtable,
                                            verbose=True)
 
+    except:
+        print('Update row failed')
+        pass
+    
+    try:
         # Add 1D spectra
         files = glob.glob('{0}_{1:05d}*1D.fits'.format(root, id))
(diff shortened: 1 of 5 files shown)


gbrammer:
Here is an example from PRIMER UDS association ``j021728m0510_unknown-f444w_00032`` at ``(α,δ) = 34.4353582, -5.1630658``.
Screenshot 2023-03-06 at 13 18 17

Issue #154, opened 2023-03-31
KeyError in Grizli-Pipeline?


flairue:
Hi. I'm following the Grizli-Pipeline notebook, using the default files. I'm getting a KeyError during the autoscript.preprocess() step.

> ---------------------------------------------------------------------------
> KeyError                                  Traceback (most recent call last)
> Cell In[19], line 7
>       3 os.chdir(os.path.join(HOME_PATH, root, 'Prep'))
>       5 # This script will do all the preprocessing of the grism *and* imaging visits 
>       6 # found in your archive query.
> ----> 7 auto_script.preprocess(field_root=root, HOME_PATH=HOME_PATH,
>       8                        visit_prep_args=visit_prep_args, **preprocess_args)    
> 
> File /opt/anaconda3/envs/spiderweb/lib/python3.11/site-packages/grizli/pipeline/auto_script.py:1960, in preprocess(field_root, HOME_PATH, PERSIST_PATH, min_overlap, make_combined, catalogs, use_visit, master_radec, parent_radec, use_first_radec, skip_imaging, clean, skip_single_optical_visits, visit_prep_args, persistence_args)
>    1956 print('\n\n\n{0} radec: {1}\n\n\n'.format(direct['product'], radec))
>    1958 ###########################
>    1959 # Preprocessing script, background subtraction, etc.
> -> 1960 status = prep.process_direct_grism_visit(direct=direct, grism=grism,
>    1961                     radec=radec, skip_direct=False, **visit_prep_args)
>    1963 ###################################
>    1964 # Persistence Masking
>    1965 for file in direct['files']+grism['files']:
> 
[...]

PR #155, opened 2023-03-31
Turn off "fix_stars" by default because `fix_star_centers` is currently broken

gbrammer:
Probably broken due to the way the pipeline creates the final mosaic with the newer functions in `grizli.aws.visit_processor`, but the correction wasn't optimal to begin with.


gbrammer:
Test failure is from `jwst`, so merging for now.

gbrammer:
Fixes https://github.com/gbrammer/grizli/issues/154.

PR #158, opened 2023-05-05
Update default to `jwst_1084.pmap` with new NIRCam flats

gbrammer:
The NIRCam flats pushed with [jwst_1084.map](https://jwst-crds.stsci.edu/context_table/jwst_1084.pmap) seem to now be comparable or better than the [`grizli` skyflats](https://s3.amazonaws.com/grizli-v2/NircamFlats2023/Best/index.html).   This update sets the default CRDS_CONTEXT  to use them in `jwst_utils`.

However, there appear to be some unmasked bad pixels, so this also includes some options to update the exposure-level DQ mask to remove pixels where the flat is below or above some percentile thresholds.  The masking also includes a "+" filter to the pixels flagged in the "mask" reference file.  For example, this flags pixels that may have charge leaked from hot pixels from the IPC effect.

### DIFF
--- a/grizli/jwst_utils.py
+++ b/grizli/jwst_utils.py
@@ -19,7 +19,8 @@ from . import GRIZLI_PATH
 QUIET_LEVEL = logging.INFO
 
 # CRDS_CONTEXT = 'jwst_0942.pmap' # July 29, 2022 with updated NIRCAM ZPs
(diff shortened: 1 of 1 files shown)


gbrammer:
Killing this, which is deprecated by https://github.com/gbrammer/grizli/pull/255.

PR #164, opened 2023-06-22
Try to use NIRCAM v8 config files

gbrammer:
Use updated config files provided by N. Pirzkal.  Public distribution TBD.


gbrammer:
This will break for extractions from `GrismFLT` that were generated from the earlier version, e.g., FRESCO on AWS Lambda.

PR #165, opened 2023-06-30
Implement additional mosaic weight strategies

gbrammer:
A number of JWST collaborations have been identifying similar issues with the stellar curves of growth in image mosaics produced with the `grizli` pipeline, i.e., that the short wavelength CoG looked too broad in the `grizli` mosaics compared to others made with the JWST pipeline.  I've come across a couple of points that were quite surprising and still a bit mysterious.

1. The first is that the "tweakshifts" alignment I'm doing in grizli for the short wavelength filters may not be as precise as assumed.  These shifts are occasionally of order 0.5 pix, but the alignment reports error-of-the-mean RMS values of 0.1-0.2 pix.  I had thought that was standard deviation ~0.1 pix, which would be much smaller with sqrt(N).  But in fact, at least for the F115W visit I checked the scatter in the alignment sources was truly of order 1 pix, or 0.1 pix on the mean.  I'm not quite sure why they could be so large, though perhaps could be related to PSF undersampling at the bluest wavelengths.  These alignment errors could have an effect on the derived stellar curves of growth if some fraction of the central bright pixels of stars were being flagged as CRs, though perhaps not necessarily in all exposures.

2. The second effect is more insidious and summarized in the plot below, which shows curves of growth measured in a) a mosaic with the nominal tweakshifts, b) a recomputed mosaic with the shift refinement turned off and c) all 16 individual exposures (from a F115W visit in UNCOVER).  The difference in the mosaics can perhaps be explained by the pixel flagging issues.  **A key point is that always looking at the curve of growth to diagnose the image quality could miss the key effect in that the main differences are in the central bright pixel or two, not necessarily in differences in the overall profile shape at larger radii.**  The mystery is in how the curves are quite different for the CoG in the individual exposures, where the central pixels are as much as 2-3x brighter than they are in the mosaics.  Since they contain a large fraction of the light, that then manifests as big differences in the overall normalization of the CoG, though the profiles at radii larger than the central few pixels are quite similar (the bottom panel is the diff of the CoG plotted above, measured with `sep`.

3. I'm not yet sure of the explanation for this latter effect, though I suspect it's related to the weighting used in the mosaic drizzling.  To make the mosaics, `grizli` converts the ERR extensions of the individual exposures to inverse variance weights, and these include a Poisson term for the detected photoelectrons, which can be quite high in the central bright pixels.  The JWST pipeline handles the variance components in a different way, which could explain why the differences are notable between my mosaics and others derived from the pipeline.

4. More tests TBD, but my suspicion is that these effects probably can explain a number of the strange CoG differences people have been finding, and also how some of them have brightness and filter dependencies.
uncover_star_cog

### DIFF
--- a/grizli/aws/aws_drizzler.py
+++ b/grizli/aws/aws_drizzler.py
@@ -782,7 +782,14 @@ def drizzle_images(label='macs0647-jd1', ra=101.9822125, dec=70.24326667, pixsca
         else:
             clean_i = remove
 
-        status = utils.drizzle_from_visit(visits[0], h, pixfrac=pixfrac, kernel=kernel, clean=clean_i, include_saturated=include_saturated, skip=skip, dryrun=dryrun)
+        status = utils.drizzle_from_visit(visits[0],
+                                          h,
+                                          pixfrac=pixfrac, kernel=kernel, 
+                                          clean=clean_i, 
+                                          include_saturated=include_saturated,
+                                          weight_type='median_err',
+                                          skip=skip,
+                                          dryrun=dryrun)
 
         if dryrun:
             filt_dict[filt] = status
--- a/grizli/aws/visit_processor.py
+++ b/grizli/aws/visit_processor.py
@@ -1716,7 +1716,7 @@ def query_exposures(ra=53.16, dec=-27.79, size=1., pixel_scale=0.1, theta=0,  fi
     return header, wcs, SQL, res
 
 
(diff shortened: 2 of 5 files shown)


gbrammer:
This PR implements a scheme to create the image mosaics with the *median* weights of the input exposure, rather than the full pixel-level weights from the `ERR` extensions of the exposure files.  This does appear to mostly fix the issues described above, as shown in the plot below where a new mosaic was created with, e.g.,  `grizli.aws.visit_processor.cutout_mosaic(..., weight_type='median_err')`.

download-2

Issue #172, opened 2023-09-12
Bug in auto_script.visit_dict_from_strings with reading footprints


jensmelinder:
Grizli version: 1.9.5

Bug appears when running auto_script.make_combined_mosaic on NIRISS direct images (F115W, F150W, F200W).

It seems that when the visit file is read, the 'footprints' item is populated by a list of lists rather than a list of footprints. This causes the call to set
`filter_groups[filt]['footprint'] = fp.buffer(0)`
(on line 4364) to fail since fp is a list in this case.

Digging through the steps I think the problem is in `visit_dict_from_strings` on line 1503.
Replacing:
```
if 'footprints' in v:
        newv['footprints'] = [utils.SRegion(fp).shapely
                              for fp in v['footprints']]
```
with
```
if 'footprints' in v:
        newv['footprints'] = [utils.SRegion(fp).shapely[0]
                              for fp in v['footprints']]
```
seems to fix the issue.

gbrammer:
Thanks @jensmelinder.  I think this should be fixed with the commit below, which adjusts how the footprints are used rather than clipping all but the first in a list, since there could potentially be multiple elements in a `utils.SRegion(fp).shapely` list.

commit https://github.com/gbrammer/grizli/pull/173/commits/0f75fe390489137ca5339780def9ca71391d58ca

PR https://github.com/gbrammer/grizli/pull/173

PR #173, opened 2023-09-13
Fix footprints bug

gbrammer:
Fix bug with legacy code that parses visit `footprints` to calculate overlaps.


gbrammer:
Addresses https://github.com/gbrammer/grizli/issues/172

PR #179, opened 2023-10-27
Fixes to pure parallel WCS

gbrammer:
The header WCS of pure-parallel exposures doesn't seem to fully reflect the dithering of the prime exposures.  These updates implement a query to the MAST database to try to identify the *prime* exposure associated with a given pure-parallel exposure, and then use the pointing of the prime to update the header WCS of the parallel.

For a particular pure-parallel exposure ``file``, e.g., `jw03383335001_02201_00001_nis_rate.fits`:

1. Find the prime exposure from the MAST query that is closest in ``EXPSTART``to ``file``
2. Use the ``apername, ra, dec`` values of the prime exposure from the MAST query and ``PA_V3`` from the ``file`` header to set the the pointing attitude with `pysiaf`
    - Assume the MAST `ra` and `dec` are at the `V2Ref` and `V3Ref` positions of the prime aperture
4. Compute the sky position of the ``CRPIX`` reference pixel of ``file`` with `pysiaf` and put that position in the ``CRVAL`` keywords


gbrammer:
The updates for a particular exposure are computed in `grizli.jwst_utils.update_pure_parallel_wcs`, which is run automatically at the end of `grizli.jwst_utils.img_with_wcs`, e.g., with `grizli.prep.fresh_flt_file`.

gbrammer:
A few more fixes to make this more robust.  

**NB:** It still doesn't work with NIRSpec `NRS_FULL_MSA` as prime because the MAST `s_region` for the `cal.fits` files is for a single slitlet and not the full aperture.

gbrammer:
The algorithm was changed in 424308d6c47d978128f48230723bfa88540af381 to query the FGS logs directly to set the pointing attitude, rather than generate it based on the prime science exposure.

PR #181, opened 2023-11-15
Update fitting.py

PJ-Watson:
If prospector is not installed, full_sed_plot() passes the SpectrumTemplate object to ax.plot(), rather than the .flux attribute.


gbrammer:
Thanks @PJ-Watson!

Issue #117, opened 2022-11-12
Value error while using auto_script.make_filter_combintions


Rimi98:
I'm getting this error while trying to make the filter combination of cal.fits files downloaded from stsci.edu using auto_script.py routine of Grizli. It is giving the value error saying there is some non ASCII characters in fits header. But I have processed images having same fits header earlier using this same routine as mentioned in ceers_nircam notebook. Its throwing this same error while trying to make catalogue from CEERS published image in their official website  [https://ceers.github.io/dr05.html?s=03#nircam-imaging](url)

the error:

auto_script.make_filter_combinations(root, weight_fnu=True, min_count=1,
        filter_combinations={'lw': ['F277W','F356W','F444W']})

indef-f277w_drz_sci.fits f277w lw block_reduce=False scl=0.24861448760009672 scl_wht=3.9234162936476285
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [74], line 1
----> 1 auto_script.make_filter_combinations(root, weight_fnu=True, min_count=1,
      2         filter_combinations={'lw': ['F277W','F356W','F444W']})

File ~/grizli/grizli/pipeline/auto_script.py:4196, in make_filter_combinations(root, weight_fnu, filter_combinations, force_photfnu, min_count, block_filters)
   4194 else:
   4195     for k in im_i[0].header:
-> 4196         head[band][k] = im_i[0].header[k]
   4198     for k in ref_h_i:
   4199         head[band][k] = ref_h_i[k]

File ~/anaconda3/envs/grizli39/lib/python3.9/site-packages/astropy/io/fits/header.py:207, in Header.__setitem__(self, key, value)
    203         self._modified = True
    204 else:
    205     # If we get an IndexError that should be raised; we don't allow
    206     # assignment to non-existing indices
--> 207     self._update((key, value, comment))

File ~/anaconda3/envs/grizli39/lib/python3.9/site-packages/astropy/io/fits/header.py:1696, in Header._update(self, card)
   1694         self._modified = True
   1695 elif keyword in Card._commentary_keywords:
-> 1696     cards = self._splitcommentary(keyword, value)
   1697     if keyword in self._keyword_indices:
   1698         # Append after the last keyword of the same type
   1699         idx = self.index(keyword, start=len(self) - 1, stop=-1)

[...]

PR #182, opened 2023-11-27
lab/niriss fixes

lboogaard:
Dear Gabe,
Here is a long overdue PR with some fixes needed to run NIRISS data
- don't open local CRDS files in update mode (fails if they are read only, as on our servers)
- skip failing header verification (workaround for #117)
- fix `n-clear` to `clear` because it won't work otherwise

FYI: @jdavies-st @rameyer @TheSkyentist

### DIFF
--- a/grizli/pipeline/auto_script.py
+++ b/grizli/pipeline/auto_script.py
@@ -3073,7 +3073,7 @@ def load_GroupFLT(field_root='j142724+334246', PREP_PATH='../Prep', force_ref=No
         for filt in ['F090W', 'F115W', 'F150W', 'F200W']:
             #key = f'{gr.lower()}-{filt.lower()}'
             # key = filt.lower()
-            key = filt.lower() + 'n-clear'
+            key = filt.lower() + '-clear'
 
             if key in masks:
                 masks[key][0] |= ((info['PUPIL'] == filt) & 
--- a/grizli/pipeline/auto_script.py
+++ b/grizli/pipeline/auto_script.py
@@ -4572,8 +4572,7 @@ def make_filter_combinations(root, weight_fnu=2, filter_combinations=FILTER_COMB
             head[band] = im_i[0].header.copy()
         else:
             for k in im_i[0].header:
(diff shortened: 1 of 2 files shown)


TheSkyentist (commented):
For some reason this also works with `key = filt.lower()` and I can't understand why this works as well.

grizli/pipeline/auto_script.py
```diff
@@ -4572,7 +4572,8 @@ def make_filter_combinations(root, weight_fnu=2, filter_combinations=FILTER_COMB
             head[band] = im_i[0].header.copy()
         else:
             for k in im_i[0].header:
-                head[band][k] = im_i[0].header[k]
+                #head[band][k] = im_i[0].header[k]
```
TheSkyentist, line 4575: You can also just skip the offending keys:
`if k == '' or k == 'HISTORY': continue`

lboogaard: sounds good, I can change that

lboogaard:
Note: https://github.com/gbrammer/grizli/blob/3b3f09da271800163443c0d09091a35a1aa6d81f/grizli/aws/visit_processor.py#L2192

gbrammer:
Adding that `n` to the NIRISS filters is a bit of an ugly hack, but it's assumed to be set as such in various places.  Does this change affect that or should it be back compatible?

gbrammer:
And for the background image, maybe you can rather write the updated version to the working directory and update the returned `bkg_file` path accordingly.

lboogaard:
Thanks - Raphael and I are looking into this

gbrammer:
Is this PR now superseded by the others @TheSkyentist submitted today?

TheSkyentist:
I believe so but I'll let @lboogaard be the final decider. The "n-clear" issue still exists (see #191) though I don't think this PR presents a global solution for combined NIRCam/NIRISS data.

lboogaard:
I agree - let's close this in favor of #191

PR #186, opened 2023-12-09
No CRDS write

TheSkyentist:
As discussed in #182, grizli attempts to write to a CRDS file, which can cause issues on systems with shared CRDS files that are write protected. The proposed solution was to copy over the file to the local CONF. 

However this does not work due to a big where the entire file path is `.lower()`ed, rather than just the filename (and so only works on case-insensitive filesystems). I fix the issue by only `.lower()`ing the filename, not the entire path (a470bff8b94a2ce499e64dce02e173266d04b0ec). 

To solve the original issue I perform a check to see if the CRDS file is writeable. If not, the CRDS file if copied to the local CONF directory and the changes are applied there (perhaps this should be the default behaviour). Any subsequent processing can then use the writeable version that is created (6d45ab6629bcaac6115403c8b2b0f07d7ac84f9d).

Finally, in the case where the original CRDS file is updated, we must check if the local CONF version is up to date. If out of date, we redo the copying step (97ff0fcd44d9caeede896768e002a0f5c5c111b6 and f49f6eed1c56e103315e7b0f1df4ef982190f416).

These fixes were developed after working through this with @lboogaard with input from @rameyer and @jdavies-st.

### DIFF
--- a/grizli/prep.py
+++ b/grizli/prep.py
@@ -6196,18 +6196,6 @@ def get_jwst_wfssbkg_file(file, valid_flat=[0.6, 1.3], make_figure=False):
     from jwst.wfss_contam import WfssContamStep
     from jwst.flatfield import FlatFieldStep
 
-    
-    with pyfits.open(file) as im:
-        if im[0].header['INSTRUME'] == 'NIRISS':
-            # Test local file
-            local_bkg_file = "nis-{0}-{1}_skyflat.fits"
-            local_bkg_file = local_bkg_file.format(im[0].header['PUPIL'],im[0].header['FILTER'])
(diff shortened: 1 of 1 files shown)


gbrammer:
Could you add a test for the [CRDS_READONLY_CACHE](https://hst-crds.stsci.edu/static/users_guide/environment.html#miscellaneous-variables) presence of the environment variable and also do the thing to write the file in CONF if it's found and True?

TheSkyentist:
Good idea! Now implemented in f9b86c3bbd52b12d97df2cebdb913cb23330a67a.

gbrammer:
Looks good now.  Thanks!

PR #187, opened 2023-12-11
Skip bad header keywords when making filter combos

TheSkyentist:
This is a fix for #117 which was briefly mentioned in #182. 
The bug only arises when multiple bands are combined to create detection images.

### DIFF
--- a/grizli/pipeline/auto_script.py
+++ b/grizli/pipeline/auto_script.py
@@ -4572,7 +4572,8 @@ def make_filter_combinations(root, weight_fnu=2, filter_combinations=FILTER_COMB
             head[band] = im_i[0].header.copy()
         else:
             for k in im_i[0].header:
-                if k == '' or k == 'HISTORY': continue
+                if k in ['', 'HISTORY']:
+                    continue
                 head[band][k] = im_i[0].header[k]
                 
             for k in ref_h_i:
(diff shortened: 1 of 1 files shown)


TheSkyentist:
These keywords are included in the headers if the drizzled images are made with `pipeline.auto_script.make_combined_mosaics` and not `aws.visit_processor.cutout_mosaic`. So this doesn't affect mosaics built with `visit_processor`. The solution is either to skip the offending keywords as suggested here, or to rewrite the portions of `make_combined_mosaics` that include these header keywords to begin with.

gbrammer:
I think it's fine to do the skip here, I was just requesting a slight rewording of the `if` test.

TheSkyentist:
Yes, in this commit I'm only skipping the offending keywords which turn out to be either an empty string or HISTORY, rather than skipping the entire header assignment as was done in the previous PR. Is there still something to be improved?

gbrammer:
Please see the tiny review request above for a little bit more flexible code which is otherwise identical to your update.

TheSkyentist:
Oh I think only write-access members of a repository can see review requests, I'm not seeing anything popping up on my end. Perhaps I'm looking in the wrong place?

gbrammer:
That's strange.  I would have thought you could see the code review as the person who originally filed the PR.  I edited the file myself with the requested change.

PR #188, opened 2023-12-12
Finished ToDo and Updated Group Visits

TheSkyentist:
I kept getting bugs when updating visit files with new exposure footprints (`auto_script.get_visit_exposure_footprints()`) after preprocessing. Eventually I tracked the bug down to the ToDo present in the `auto_script.py` file. I finished the ToDo and no longer get the error. Hopefully this is the intended fix.


gbrammer:
Looks good.  Thanks!

PR #189, opened 2023-12-12
Remove requirement to install AWS dependency

TheSkyentist:
When running NIRISS JWST data, `grizli` usually requires that `awscl` and `boto3` be installed. E.g. one must install using `pip install grizli[jwst,aws]`. I believe I have removed these requirements if AWS is **not** being used. In addition, I have suppressed messages that notify the use that `awscli` is not installed when used to test if the `awscli` is installed or not.


gbrammer:
Nice to finally silence that shell warning!

PR #203, opened 2024-01-27
Update filter lists to enable NIRISS F090W WFSS processing

gbrammer:
Add ``F090W`` to filter lists so that it is included as an available filter for NIRISS WFSS processing.

### DIFF
--- a/grizli/grismconf.py
+++ b/grizli/grismconf.py
@@ -1341,6 +1341,9 @@ def load_grism_config(conf_file, warnings=True):
      """
             msg = f' ! Scale NIRISS sensitivity by {hack_niriss:.3f} prelim flux correction'
             utils.log_comment(utils.LOGFILE, msg, verbose=warnings)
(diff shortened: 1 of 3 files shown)


gbrammer:
@jkmatharu

gbrammer:
Here are some preliminary sky background images for NIS F090W made from the [COM-1089](https://www.stsci.edu/cgi-bin/get-proposal-info?id=1089&observatory=JWST) wavecal calibration data.  They are essentially smoothed versions of a few individual exposures.  There weren't enough dither positions to efficiently mask all source spectra, but these files do provide a significant improvement over the current files in CRDS.

```
nis-f090w-gr150c_skyflat.fits
nis-f090w-gr150r_skyflat.fits
```

[nis-f090w-grism-sky-v0.zip](https://github.com/gbrammer/grizli/files/14229034/nis-f090w-grism-sky-v0.zip)

PR #204, opened 2024-02-06
Corrected language

TheSkyentist:
Found a typo, the intermediate saved files should be called "wfssback" OWTTE, rather than "skyflat".

### DIFF
--- a/grizli/prep.py
+++ b/grizli/prep.py
@@ -6256,7 +6256,7 @@ def get_jwst_wfssbkg_file(file, valid_flat=[0.6, 1.3], make_figure=False):
             return bkg_file
 
     # Get local background file location
-    local_bkg_file = f"nis-{h['PUPIL']}-{h['FILTER']}_skyflat.fits".lower()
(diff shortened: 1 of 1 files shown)


gbrammer:
I have made my own versions of those files with the `skyflat` extension (see https://github.com/gbrammer/grizli/pull/203). I'm worried that function is getting a little convoluted with generating the filenames and certain assumptions should be generalized to modifiable keyword arguments.

TheSkyentist:
Whoops, that's a good catch, I wouldn't want to break anyone's workflow. 
Perhaps it's just better to leave it as is right now. Or perhaps it's worth distributing your WFSS Backgrounds with grizli? Have these been made available? I'd definitely like to check them out. 

Regardless, I can close this PR.

gbrammer:
I haven't done that systematically yet for all filters, but would be good to do. The CRDS sky backgrounds are a reasonable guess to get started, but pretty clearly leave some room for improvement.

PR #206, opened 2024-02-06
Updating to Polynomial

TheSkyentist:
As of [SciPy 1.12.0](https://docs.scipy.org/doc/scipy//release/1.12.0-notes.html#id1) `polyfit` and `polyval` can no longer be directly imported from SciPy. This caused import errors when running grizli. I suspect this is due to the fact that `polyfit` and `polyval` are part of the old polynomial API, which has been upgraded with [NumPy Polynomial](https://numpy.org/doc/stable/reference/routines.polynomials.html) package, primarily resulting in greater numerical stability. While I could have just fixed the imports, I went through and updated the code to use the new Polynomial API. If preferred, I can simply replace the scipy imports with the correct/not broken numpy imports.

### DIFF
--- a/grizli/fitting.py
+++ b/grizli/fitting.py
@@ -2513,7 +2513,7 @@ class GroupFitter(object):
             
             
         """
-        from numpy.polynomial.polynomial import polyfit, polyval
+        from numpy.polynomial import Polynomial
         from scipy.stats import t as student_t
         from scipy.special import huber
         import peakutils
@@ -2676,9 +2676,9 @@ class GroupFitter(object):
             zgrid_zoom = []
             for ix in indexes[:max_peaks]:
                 if (ix > 0) & (ix < len(chi2)-1):
-                    c = polyfit(zgrid[ix-1:ix+2], chi2[ix-1:ix+2], 2)
-                    zi = -c[1]/(2*c[0])
-                    chi_i = polyval(c, zi)
+                    p = Polynomial.fit(zgrid[ix-1:ix+2], chi2[ix-1:ix+2], deg=2)
+                    zi = p.deriv().roots()[0]
+                    chi_i = p(zi)
                     
                     # Just use grid value
                     zi = zgrid[ix]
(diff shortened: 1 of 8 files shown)


gbrammer:
I like the modernization idea, but the wholesale changes to the new `Polynomial` behavior are a bit risky, particularly in the way the order of the polynomial coefficients are interpreted is exactly reversed!  Those `coeffs[::-1]` throughout aren't pretty, but at least everything (sort of) works!  

Could we just switch to `from numpy.polynomial.polynomial import polyfit, polyval` for now and file switching to the `Polynomial` objects as a WIP Issue?

TheSkyentist:
Sounds good, I've submitted PR #207 to address the short-term issues, but will continue testing against this PR and keep this updated.

TheSkyentist:
This PR now passes all CI checks and has been tested on real data to ensure it gives the same results.

gbrammer:
Thanks for keeping an eye on this.  Can you update minimum versions on the `numpy` and `scipy` dependencies in [setup.cfg](https://github.com/gbrammer/grizli/blob/master/setup.cfg) to ensure compatibility with these changes (e.g., `cumtrapz` > `cumulative_trapezoid`)?  I added `scipy<1.14` the other day to fix a bug in the external (and apparently stale) `peakutils` module, which uses `scipy.integrate.simps` that has been renamed to `simpson`.

TheSkyentist:
This is a good point, we actually maintain greater compatibility with this change. The "new" Polynomial syntax has been around in NumPy since 1.4, which is over 10 years old at this point. It's true that scipy has finally removed the deprecated "simps/cumtrapz" syntax, I've forked [peakutils](https://github.com/TheSkyentist/peakutils) and made the necessary updates, I wonder if we can pull from there for now. I wonder if there is a more modern package with the same functionality.

TheSkyentist:
I've just pushed the change requiring NumPy >=1.4. There are likely other parts of grizli that would require newer NumPy versions, but this is a start.

gbrammer:
Thanks. Can you find the minimum scipy needed for cumulative_trapezoid?

TheSkyentist:
Scipy 1.6 introduced the new scipy.integrate.

gbrammer:
Merged!  Thanks again, @TheSkyentist.

PR #207, opened 2024-02-11
Import polyfit/val from numpy

TheSkyentist:
Import polyval from numpy instead of scipy to fix mitigate issues.

### DIFF
--- a/grizli/fitting.py
+++ b/grizli/fitting.py
@@ -2513,7 +2513,7 @@ class GroupFitter(object):
             
             
         """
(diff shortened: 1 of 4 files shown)


gbrammer:
I think it would be safest to

`from numpy.polynomial.polynomial import polyfit, polyval`

rather than from the general `numpy` namespace  with `from numpy import polyfit, polyval`.

TheSkyentist:
Good idea, I've committed the requested changes.

gbrammer:
Thanks!

Issue #190, opened 2023-12-12
Crash if no Direct in same filter as Grism for NIRISS (possibly other instruments too)


TheSkyentist:
I am working on a NIRISS program with F200W direct and GR150R grism in F200W, F150W, and F115W. 

During `grism_prep`, the code will fail if there are no Direct observations in the same filter as the Grism observations:

This is due to the following:

1) When loading the GroupFLT the following code snippet is run:

https://github.com/gbrammer/grizli/blob/3b3f09da271800163443c0d09091a35a1aa6d81f/grizli/pipeline/auto_script.py#L3071C1-L3084C40

which checks if there is an observation matching both the grism and pupil information. For my program, this will be true for F200W, F150W and F115W. 

2) Then, the reference filter is decided based purely on the previous information:

https://github.com/gbrammer/grizli/blob/3b3f09da271800163443c0d09091a35a1aa6d81f/grizli/pipeline/auto_script.py#L3071C1-L3084C40

The result is that the grism looks for a detection image in each of my grism bands, which causes an error:

https://github.com/gbrammer/grizli/blob/3b3f09da271800163443c0d09091a35a1aa6d81f/grizli/pipeline/auto_script.py#L3148C36-L3148C36

The preferred behavior would be the following:
1) Use the matching direct filter for the grism if it exists. 
2) If it does not, specify the priority of direct images to use for grism-filter pairs. 
3) At the very least, the order of direct filters should be used from the Grism Reference Filters dictionary.

This is especially relevant for pure parallel data where the grism/direct filter combos can be a grab bag. 
I don't believe this affects NIRCam after taking a cursory look, but it may have a similar affect.

gbrammer:
I think I'd prefer that the user have the responsibility for assigning the direct image reference for these cases.  

You could do this either by simply making a symbolic link of the file with the expected filename to some other one that actually exists.  Or, alternatively you can explicitly specify the direct image reference with the `force_ref` keyword in [pipeline.auto_script.grism_prep](https://github.com/gbrammer/grizli/blob/master/grizli/pipeline/auto_script.py#L3250).

TheSkyentist:
I agree that the user should specify what the fallback is. However, I thought that was the role Grism Reference Filters dictionary. If that's not the case, perhaps we can patch in a more appropriate error warning statement?

TheSkyentist:
[...]

PR #210, opened 2024-02-23
Added comments on failure

TheSkyentist:
Added logging messages if a reference image or segmentation image can't be found. In reference to #190.


TheSkyentist:
Added a patch for the bug (I think?). If this was not the bug you meant to fix, can you clarify which bug?

gbrammer:
Yes, that's it.  I had started a code review and highlighted the lines, but it seems like I didn't submit the review at the end.

Looks good now.

PR #212, opened 2024-02-28
Optional 1/f when setting JWST/HST keywords

TheSkyentist:
Currently if `set_jwst_to_hst_keywords` is called  before `initialize_jwst_image` then a 1/f correction is always applied. This can even lead to 1/f correction being applied twice. It would be good to allow the user to pass an option to disable this if they'd like. Otherwise this won't change any of the default behaviour.

### DIFF
--- a/grizli/jwst_utils.py
+++ b/grizli/jwst_utils.py
@@ -1456,7 +1456,7 @@ def initialize_jwst_image(filename, verbose=True, max_dq_bit=14, orig_keys=ORIG_
 #                  'F480M': [1.879639e-21, 1.453752e-30, 4.8152]}
 # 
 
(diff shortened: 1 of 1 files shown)


gbrammer:
This fix seems fine and relatively harmless, but it could be more automatic to also have [exposure_oneoverf_correction](https://github.com/gbrammer/grizli/blob/7da640e68a83f93e1f58802c1553ea33d7184233/grizli/jwst_utils.py#L926) first check for the presence of the `ONEFEXP` keyword in the file and then break out without doing anything if the keyword is found.  And add an additional keyword option ``force`` that defaults to ``False`` in `exposure_oneoverf_correction` but that will redo the correction if True even if it had already been done previously.

TheSkyentist:
This was a good suggestion. I've added the check against the ONEFEXP keyword, as well as adding a force option in the function.

Issue #221, opened 2024-05-03
ROOT_PATH set automatically when importing visit_processor


TheSkyentist:
ROOT_PATH, which controls the directory of (most) execution is set when the `visit_processor` module is imported:

https://github.com/gbrammer/grizli/blob/1ed492f4395add1cbcaa727f4b7f3b242e80979b/grizli/aws/visit_processor.py#L18-L20

However, it would be helpful to allow the user to (optionally) pass this as a parameter when calling functions from `visit_processor`.  

This would apply to the following functions in `visit_processor`:

1. `get_assoc_yaml_from_s3`
2. `check_jwst_assoc_guiding`
3. `process_visit`
4. `get_wcs_guess`
5. `run_one`

TheSkyentist:
Closed due to reasons outlined in the comments to #222

PR #222, opened 2024-05-03
Added ability to specify ROOT_PATH in visit_processor

TheSkyentist:
Addresses #221 by allowing the user to specify ROOT_PATH for functions within `visit_processor`.

### DIFF
--- a/grizli/aws/visit_processor.py
+++ b/grizli/aws/visit_processor.py
@@ -957,7 +957,7 @@ def update_visit_results():
     pass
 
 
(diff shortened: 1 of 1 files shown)


gbrammer:
This is set as a global variable `visit_processor.ROOT_PATH`.  You should get the same behavior as the proposed changes by setting it explicitly:
```python
from grizli.aws import visit_processor
visit_processor.ROOT_PATH = '/my/root/path'
```

It doesn't necessarily have to be set immediately after importing, but can rather be done / changed at any time.

TheSkyentist:
This is good to know! I'll close this PR, as the required behavior can be achieved with existing means.

PR #232, opened 2024-06-26
 Compliance with NumPy 2.0

TheSkyentist:
NumPy 2.0.0 has been released, removing np.cast. This should fix compliance with NumPy.
This is won't break `grizli["jwst"]` for now as the JWST pipeline is pinned to NumPy < 2.0.0 but this may have current consequences for HST reduction, and will eventually affect JWST when they upgrade to newer NumPy version.


gbrammer:
thanks @TheSkyentist !

Issue #235, opened 2024-06-27
Replace PeakUtils?


TheSkyentist:
Now that peakutils is incompatible with the newest SciPy versions and is no longer maintained, should it be replaced with an equivalent? Is `scipy.signal.find_peaks` a viable alternative?

gbrammer:
Yes, that sounds like a good idea. The peakutils bits aren't particularly sophisticated, and are probably easily replaced with something like you mention.

TheSkyentist:
It seems as though this Peakutils functionality was ported to scipy in
2017.  I’ll make sure the behavior is equivalent and make a PR with the
change.
https://github.com/scipy/scipy/issues/7393

gbrammer:
Cool. If it's actually even the same algorithm that would be ideal. One useful bit I use from peakutils is the way it can find peaks with a user-specified minimum separation between them.

TheSkyentist:
After some investigating and testing I can report that the algorithms are the same! PeakUtils does some internal rescaling of the data, but we can easily do that in-situ. The conversion between the two is straightforward:

```python
# PeakUtils Implementation
indices = peakutils.indexes(
        y,
        thres=t,
        min_dist=d
    )

# SciPy Signal Implementation
indices,_ = scipy.signal.find_peaks(
        y,
        height=t*(y.max()-y.min())+y.min(), # Rescale
        distance=d+1 # Note the off by one here!
    ) 
```
I'll submit a PR with the required changes.

TheSkyentist:
Closed by #236

PR #236, opened 2024-06-28
Switch from PeakUtils to Scipy.Signal

TheSkyentist:
Fix addressing #235.

### DIFF
--- a/grizli/tests/test_utils.py
+++ b/grizli/tests/test_utils.py
@@ -262,3 +262,11 @@ class UtilsTester(unittest.TestCase):
         print('sip_rot90 tests passed')
         
         plt.close('all')
+
+    def test_find_peaks(self):
+
+        "Test Find-Peaks function in utils"
+        
+        signal = np.array([0, 1, 0, 2, 1, 0, 3, 0])
+        expected_peaks = np.array([3, 6])
+        np.testing.assert_array_equal(utils.find_peaks(signal), expected_peaks)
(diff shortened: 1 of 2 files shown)


gbrammer:
Thanks, looks pretty straightforward.  It looks like you figured out the mapping between the two modules with the thresholding.  Rather than overloading the `height=` keyword with some array transformations, could you consider writing a function (and a test) like `utils.threshold_peak_indexes` or something that takes "threshold" and "min_separation" keywords and behaves like `peakutils.indexes`?

TheSkyentist:
I'd be willing to do this, but perhaps this would be quite a short function in utils. Perhaps I can clean up the creation of the height value to make it clearer. Let me make some changes and let me know if they make things more readable or if you'd prefer to make a new function.

gbrammer:
There's nothing wrong with a short function!  Just to make the behavior explicit and easily testable.

TheSkyentist:
Okay I'll refactor in a bit.
Also a note that the CI is failing because the latest photutils fails with
the latest drizzlepac (as of 3 hours ago). Drizzlepac has changes in place
for their next release, but for now we might want to pin photutils as well.

gbrammer:
OK, this looks great.  Can you add a test to `grizli/tests/test_utils.py` for the new function?  The test doesn't have to be super comprehensive, just something that verifies that the function works as expected given `numpy`, `scipy`, etc.

gbrammer (approved):
Looks good!

PR #239, opened 2024-07-20
Support for Context Maps

TheSkyentist:
The current grizli implementation of AstroDrizzle was not correctly computing context maps, since it was always setting the ID of the input image to 1, rather than incrementing it. This was fixed in 77412f6. I've also added the ability to optionally output context maps when creating mosaics for the direct and grism images.

### DIFF
--- a/grizli/aws/aws_drizzler.py
+++ b/grizli/aws/aws_drizzler.py
@@ -531,7 +531,7 @@ def segmentation_figure(label, cat, segfile):
     th.close()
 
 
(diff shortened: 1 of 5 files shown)


TheSkyentist:
A quick note, AstroDrizzle does not support more than 32 images in a drizzle. So if this happens, print a warning and continue.

gbrammer:
Finally getting around to this.  Could you rename the parameter controlling whether to output the context image `out_ctx` to `write_context`?  

And for the issue with more than 32 images, how about just wrapping the integer at 32?  Since `drizzle_array_groups` is effectively calculating the context anyway, you don't need to add logic whether or not to do it, `drizzle_array_groups` can have the parameter default  `drizzle_array_groups(first_uniqid=1)` and then inside `adrizzle.do_driz(..., uniqid=((first_uniqid - 1 + i) % 32) + 1)`.

Finally, it looks like there are some debugging lines that are commented out and should be removed for clarity.

Thanks!

TheSkyentist:
This is great feedback, I’ll get this implemented soon.

TheSkyentist:
I've addressed the handling of ID assignment, as well as made the warning log use the grizli msg.log syntax. In addition, I properly implemented the in-progress sections that were commented out.

gbrammer:
Looking good.  

- There seems to still be one commented block in `aws_drizzler`
- Can you rename `ctx_out` to something a bit more meaningful (e.g., `write_ctx_file`, `write_context`)

TheSkyentist:
Good suggestions, I've reformatted to write_ctx and fixed the last comment block.

TheSkyentist:
I've updated this commit to be merge-able after the most recent changes.

gbrammer (approved):

PR #229, opened 2024-06-08
formatting and docs for utils

KostasValeckas:
This is a suggestion of the discussed doc and formatting walk-through. 

Specifically: 

* added missing / completed incomplete docstrings using AI tools

* went through formatting using `black` (some changes reverted where did they did not make sense)

Any ambiguities got marked with TODO comments (only a few of those).

Any AI-generated or inspirited docstrigs are tagged with a note that discloses this.

My suggestion is that we could use this PR to discuss how the changes could be made in the best way. That said, it is made consistent in the sense that it can be merged, if my edits are in agreement of what is best for the code.

All tests pass.

PR #240, opened 2024-07-26
docs and formatting: stack.py, combine.py, catalog.py, fitting.py

KostasValeckas:
Updated docs and formatting for stack.py, combine.py amd catalog.py

This is in continuation from PR #229 as agreed with @gbrammer .

Specifically:

* added missing / completed incomplete docstrings using AI tools

* adressed some unused parameters (as this is a part of docs)

* went through formatting using black.

In `stack.py`,  one ambiguitie is marked with a TODO comment (`optimal_extract` method).

in `combine.py` the docs I made might be overkill for the small methods in `SIP_WCSMap` class - I chose to be consistent with rest of the code, but that might be too much in this case - since I know this will be reviewed I leave it for evaluation. 

In `catalog.py`, I left a few TODO comments around some parameters that might need more attention. I have made docstrings for all the wrappers for `query_tap_catalog` - If you would rather have the docstrings reference to the `query_tap_catalog` instead to make things more compact please let me know - I can easily change this.

In `fitting.py` I have adressed a handfull of unused parameters - some I have tried fixing, some I have marked with comments. In all cases, I have attached TODO comments for easier review. 

Any AI-generated or inspirited docstrigs are tagged with a note that discloses this.

All tests pass.

### DIFF
--- a/grizli/catalog.py
+++ b/grizli/catalog.py
@@ -55,7 +55,6 @@ def table_to_radec(table, output="coords.radec"):
     output : str
         Output filename. Default is `coords.radec`.
 
-    Note: Some of this documentation is AI-generated and will be reviewed.        
     """
 
     if "X_WORLD" in table.colnames:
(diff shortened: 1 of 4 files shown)


KostasValeckas:
If I am not mistaken, the failed workflow here is caused by error in package installation (Flask - I think). Tests on python 3.9 all pass locally on my machine.

gbrammer (commented):
Looks great!

Issue #242, opened 2024-08-13
Missing `objective_epsf_center' function. 


TheSkyentist:
708af1549d736d7a0a93ab58eba16fcb1ce64d94 removed `objective_epsf_center` from `utils.py`. However, this function is stilled referenced throughout the code and is used for fitting PSF centers (for example with `wfc3dash`.

PR #243, opened 2024-08-13
Added objective_epsf_center back

TheSkyentist:
Fix to #242. Identical to the function from d0c1adaf7c1517ab4d3ecb2184c5fc59c86b928a but with the commented out bits removed and the documentation lengthened to match `objective_epsf`.


gbrammer:
Can you close https://github.com/gbrammer/grizli/issues/242 if it is now addressed?

PR #249, opened 2024-10-07
Implement `weight_type="jwst_var"` strategy for mosaics with full error propagation

gbrammer:
For a weighted sum of noisy data $y = \sum_i(w_i x_i)  / \sum_i(w_i)$, the "optimal" weight that produces the minimum $\mathrm{Var}(y)$ is $w_i = 1 / \sigma_i^2$, i.e, weighting by the inverse variance of $x_i$.  For general $w_i$, 

  $\mathrm{Var}(y) = \sum_i (w_i^2 \sigma_i^2) / (\sum_i w_i)^2$. (**Eq. 1**)

This reduces to $\mathrm{Var}(y) = 1 / \sum_i w_i$ for the optimal weights $w_i = 1/\sigma_i^2$.

The "drizzle" co-addition of images is essentially the linear sum above after projecting an input pixel grid to the desired output pixel grid, accounting for camera distortions, pixel scale, etc.  The noise-optimal weights for drizzle resampling image $I$ are $W = 1 / V$, where $V$ is the variance image created from the propagated uncertainties of the noise model for image $I$, including, e.g., detector read noise and the Poisson term from the recorded photoelectrons.  However, with the somewhat undersampled space telescope images from HST and JWST, including the Poisson component tends to corrupt the *spatial* morphology of compact, high S/N sources where the Poisson term dominates.

In relatively recent updates to ``grizli``, we included implementation of a number of weighting schemes using different combinations of the exposure noise model.  The default used for the ``v7`` mosaics of the DAWN JWST Archive use ``weight_type="jwst"``, where the weights for image $i$ are calculated with

$W_i^{-1} = V_{i,\mathrm{RN}} + \mathrm{Median}(V_{i,\mathrm{Poisson}})$.  (**Eq. 2**)

This therefore includes relative weighting for, e.g., both varying exposure times and background levels within the list of exposures combined to a final mosaic, but recovering the full variance thus required calculating the Poisson term for detected sources using the science and exposure time maps.

Here we implement a new scheme ``weight_type="jwst_var"`` that produces the full variance map of the drizzle product at runtime using a combination of **Eqs. 1** and **2**:

  $V_\mathrm{drz} = \sum_i (W_i^2 V_i^2) / (\sum_i W_i)^2$,

where $V_i = V_{i,\mathrm{RN}} + V_{i,\mathrm{Poisson}}$ includes the full Poisson map.

Running
```python
grizli.aws.visit_processor.cutout_mosaic(..., weight_type="jwst_var")
```
creates an additional image product ``..._drc_var.fits`` from this combination.  The file ``..._drc_wht.fits`` is the inverse variance map as before *without* the Poisson term for discrete sources.

### DIFF
--- a/grizli/aws/aws_drizzler.py
+++ b/grizli/aws/aws_drizzler.py
@@ -782,21 +782,24 @@ def drizzle_images(label='macs0647-jd1', ra=101.9822125, dec=70.24326667, pixsca
         else:
             clean_i = remove
 
-        status = utils.drizzle_from_visit(visits[0],
-                                          h,
-                                          pixfrac=pixfrac, kernel=kernel, 
-                                          clean=clean_i, 
-                                          include_saturated=include_saturated,
-                                          weight_type='median_err',
-                                          skip=skip,
-                                          dryrun=dryrun)
+        status = utils.drizzle_from_visit(
+            visits[0],
+            h,
+            pixfrac=pixfrac,
+            kernel=kernel,
+            clean=clean_i,
+            include_saturated=include_saturated,
+            weight_type='median_err',
+            skip=skip,
+            dryrun=dryrun
+        )
 
         if dryrun:
             filt_dict[filt] = status
             continue
 
         elif status is not None:
-            sci, wht, outh, filt_dict[filt], wcs_tab = status
+            sci, wht, var, outh, filt_dict[filt], wcs_tab = status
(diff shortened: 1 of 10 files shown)


gbrammer:
Bug fixes in https://github.com/gbrammer/grizli/pull/250

PR #246, opened 2024-09-11
docs and formatting for fake_image.py, girsmconf.py, jwst_utils.py

KostasValeckas:
Updated docs and formatting for `fake_image.py`, `girsmconf.py`, `jwst_utils.py` .

This is a continuation from PR's #229 and #240 as agreed with @gbrammer .

Specifically:

* added missing / completed incomplete docstrings using AI tools.

* went through formatting using black.

* removed or edited some unused parameters

Very little done to `fake_image.py` .

For `girsmconf.py`, in the `CRDSGrismConf` class, I had issues with determining the interplay between the many DISP methods (`DISPX`, `INVDISPX`) and all alike, and the model evaluation methods that are being employed within (specially the inverse). I had made a best-attempt, and left some TODO comments on where I got stuck. Many of these methods are repetitive, so please feel free to let me know if/where the docs are incorrect (and if possible what corrections should be done) - I would happily to do another iteration and apply any feedback myself.

In `jwst_utils.py`, some parameters were never used. These were mostly keyword args to other methods. I have exchanged hardcoded keyword args with the unused parameters where the intention for this was clear. All relevant parts are marked with `TODO` comments for review.

PR #252, opened 2024-10-09
Docs and formatting of multifit.py

KostasValeckas:
Updated docs and formatting for `multifit.py`.

This is a continuation from PR's #229, #240 and #246 as agreed with @gbrammer .

Specifically:

* added missing / completed incomplete docstrings using AI tools.

* went through formatting using black.

* adressed some unused parameters

Some unused parameters are marked with TODO comments. All AI-inspired docs are marked with a note in the docstring.


KostasValeckas:
The failed test is also present in PR #255 - I suspect it is not caused by this PR (as no functional code is changed).

PR #255, opened 2024-10-14
Add updated NIRCam photometric zeropoints based on the sky background in archival exposures

gbrammer:
The flat sky background estimated from individual exposures can provide a robust *relative* correction to the NIRCam photometric zeropoints, where multiple detectors observe the same intrinsic "sky" surface brightness simultaneously.


gbrammer:
**Still to do:** add documentation

PR #258, opened 2024-11-05
Support for Context Map Planes

TheSkyentist:
In order to support more than 32 images per drizzle, we need to use Astrodrizzle planes. Here I've added support for this, but it still requires some testing.


TheSkyentist:
Rebased to ease merging.

TheSkyentist:
I noticed that in a recent version there was a log added if drizzled images exceeded 32 because context maps won't be correct. This allows the context maps to use planes so it can capture >32 images.

PR #260, opened 2024-11-11
Use environ context for tweak align

TheSkyentist:
During tweak align, grizli uses a hard-coded CRDS context to apply a photom correction. However, this is quite an old context, and if the CRDS folder is write only (as it is in shared environments), it will cause a failure. As this doesn't affect the final output (only used to compute the tweak shifts) I have added in a check to use the environment context value if it exists.

### DIFF
--- a/grizli/utils.py
+++ b/grizli/utils.py
@@ -8036,6 +8036,8 @@ def drizzle_from_visit(
         if scale_photom:
             # Scale to a particular JWST context and update header keywords
             # like PHOTFLAM, PHOTFNU
+            if 'CRDS_CONTEXT' in os.environ:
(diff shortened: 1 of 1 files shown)


PJ-Watson:
This isn't the only place that I've noticed the CRDS is essentially hard-coded - it also appears as a default parameter in `get_crds_zeropoint()` (although this seems to be a redundant function), and in `aws.visit_processor.cutout_mosaic()` (which itself also calls `utils.drizzle_from_visit()`). The latter function is a bit of a problem, as the `context` parameter isn't passed through or checked when using `prep.iterate_tweak_align()`. I can always submit a separate pull request though if that's easier!

TheSkyentist:
Rebased to make merging easier.

PR #263, opened 2024-12-19
Create wavelength_dependent_psf.ipynb

keithbuckholz:
Added notebook version of wavelength-dependent PSF grism sim.


keithbuckholz:
Apologies. This was not meant to be requested here.

PR #271, opened 2025-06-24
Fix masking parts of mosaics where only a single exposure contributes

gbrammer:
The function in `grizli.prep.find_single_image_CRs` enables masking of individual exposures in parts of the final mosaic where only a single exposure contributes.  With the previous behavior, small clumps of pixels where N-1 exposures in an association of N exposures were flagged as CRs were then excluded from the final mosaic.  For N=2, this results in many holes across the mosaic but where pixels should probably be acceptable, the primary intention of the mask function is to mask large parts of the mosaic only sampled by single exposures, e.g., around the edges of a dither pattern.

This update adds a parameter `grizli.prep.find_single_image_CRs(erode_ctx_single=3)` that performs a binary erosion / dilation on the single image mask derived from the context image to fill these holes.

An additional minor change treats the HST/ACS DQ=128 bit as valid, which further helps to fill gaps, especially in N=2 visits.

### DIFF
--- a/grizli/prep.py
+++ b/grizli/prep.py
@@ -343,7 +343,8 @@ def fresh_flt_file(
 
     if crclean and has_scrappy:
         for ext in [1, 2]:
-            print("Clean CRs with LACosmic, extension {0:d}".format(ext))
+            msg = f"Clean CRs with LACosmic, extension {ext:d}"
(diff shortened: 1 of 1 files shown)


gbrammer:
Also includes additional minor changes to message logging and removes some old large commented code blocks.

PR #279, opened 2025-11-11
Add a function that calculates trace in unrotated frame

ivastar:
This allows for a trace comparison between the traces as calculated from `jwst.datamodels` and from `grizli` in the original frame of the images, prior to rotation.


gbrammer:
Can you check the `flake` bug?

PR #285, opened 2026-03-04
Roll back to force drizzlepac 3.6.2

gbrammer:
For some reason associations processed `drizzlepac>=3.7.0` are full of masked bad pixels, and resolving this will require more investigation.  This update rolls back to `drizzlepac==3.6.2` and an older version of `jwst<1.17` that is known to work and also remains consistent with everything processed thus far in DJA.

### DIFF
--- a/grizli/jwst_utils.py
+++ b/grizli/jwst_utils.py
@@ -4774,7 +4774,8 @@ def flag_nirspec_hot_pixels(
 
     bits = get_jwst_dq_bit(jwst_dq_flags)
 
-    mask = (rate["DQ"].data & bits > 0) | (rate["ERR"].data <= 0)
+    mask = (rate["DQ"].data.astype(np.int32) & bits > 0)
+    mask |= (rate["ERR"].data <= 0)
     mask |= (rate["SCI"].data < -3 * rate["ERR"].data) | (
         ~np.isfinite(rate["SCI"].data)
     )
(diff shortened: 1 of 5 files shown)


gbrammer:
The `setuptools==67.8.0` requirement is last point at which `pkg_resources` is available.  Installation of the older `drizzlepac` version was failing on `python==3.12` with `pkg_resources` ImportErrors.

PJ-Watson:
This one I can offer some insight on, since I ran into the same issue a few months back - the bad pixel processing is due to the reversion in HLA:1061 (https://github.com/spacetelescope/drizzlepac/pull/1767). The previous behaviour essentially skips over NaN values (zero-weight pixels) in the input images, the current behaviour propagates NaNs into the final mosaic. The simplest change needed is to reimplement (re-revert?) two lines in `src/drizzlepac.c`, e.g. https://github.com/spacetelescope/drizzlepac/compare/main...PJ-Watson:drizzlepac:main , but it requires maintaining a separate fork. A more involved change would be replacing NaN values with 0 prior to drizzling, then figuring out a way to put them back in afterwards, or replacing `drizzlepac` with [`drizzle`](https://github.com/spacetelescope/drizzle), which also maintains the old behaviour.

gbrammer:
Thank you @PJ-Watson!  Glad to see even my minimalist PR comments can generate some useful feedback and discussion.

I don't know why they've set the NaNs in the images to begin with if all data quality flagging is already handled with the DQ extensions.  Maybe the flats have zeros in them or something so the NaNs aren't explicitly set but rather are induced.  Setting the NaNs to zeros in `grizli` would actually be pretty easy, so I'll look into that as a quick fix.

I'd really like to move to pure `drizzle` since `drizzlepac` brings along a lot of heavy GUI, etc., dependencies I don't need, but I still use a bunch of the bad pixel flagging from `AstroDrizzle`.  I haven't looked into it for years, but the motivation for using `drizzlepac` was for uniform processing across both HST and JWST and the datamodels and functionality in `jwst` hadn't been propagated to HST.

PR #288, opened 2026-04-10
NIRISS WFSS Background Reference Flag Update

TheSkyentist:
JWST pipeline / references changed behaviour for NIRISS WFSS Background files.
They are now referenced as bkg, not wfssbkg. 
Passing wfssbkg will result in an error.

### DIFF
--- a/grizli/prep.py
+++ b/grizli/prep.py
@@ -7819,7 +7819,7 @@ def get_jwst_wfssbkg_file(
     from jwst.wfss_contam import WfssContamStep
     from jwst.flatfield import FlatFieldStep
 
-    bkg_file = WfssContamStep().get_reference_file(file, "wfssbkg")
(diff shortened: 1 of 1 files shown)


gbrammer:
Is this update consistent with older versions of `jwst` and `crds`, i.e, `jwst<1.17` and `crds<13.0` in the `grizli` setup dependencies?  I started looking at updating things to bring us into the modern era with `jwst=2.0` and noticed a few other things had changed that I just committed some fixes elsewhere to ensure back- and fore- compatibility.

TheSkyentist:
This was a change in the `jwst` package introduced in 1.12.0 (https://github.com/spacetelescope/jwst/issues/9584). So this currently does raise an issue with `grizli` with `jwst < 1.17`.

PR #291, opened 2026-05-18
Fix dqflags import in latest version of stdatamodels.

PJ-Watson:
`stdatamodels>=5.0.0` no longer imports `pixel` within `datamodels.mask`. This corrects the import to `datamodels.dqflags`, and should be backwards compatible at least to `stdatamodels>=1.0.0` (February 2023).


gbrammer:
Thanks!

PR #292, opened 2026-05-19
Check if 1/f correction was applied using the STScI JWST pipeline.

PJ-Watson:
Check if the 1/f correction was already applied using the JWST pipeline ([`CleanFlickerNoiseStep`](https://jwst-pipeline.readthedocs.io/en/stable/api/jwst.clean_flicker_noise.clean_flicker_noise_step.CleanFlickerNoiseStep.html#jwst.clean_flicker_noise.clean_flicker_noise_step.CleanFlickerNoiseStep)) via the header keywords. As `exposure_oneoverf_correction()` is called from `jwst_utils.set_jwst_to_hst_keywords()`, at present there's no way to stop this correction being applied again.

### DIFF
--- a/grizli/jwst_utils.py
+++ b/grizli/jwst_utils.py
@@ -1445,7 +1445,10 @@ def exposure_oneoverf_correction(
 
         return None, 0
 
-    if ("ONEFEXP" in im[0].header) and im[0].header["ONEFEXP"] and (not force_oneoverf):
+    if (
+        (("ONEFEXP" in im[0].header) and im[0].header["ONEFEXP"])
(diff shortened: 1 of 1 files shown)


TheSkyentist:
Currently `set_jwst_to_hst_keywords` has an argument called `oneoverf_correction`. Settings this to false would stop 1/f correction from being applied.

PJ-Watson:
It does indeed - apologies for the ambiguity. The issue there is that `set_jwst_to_hst_keywords` is often called without arguments (or only with `reset=True`). Passing through `**kwargs` every time that function is called would be a significantly more involved change - and I'm not sure it'd be safe to change the default `oneoverf_correction=True`, since it's only recently that the pipeline applies the 1/f correction to NIRISS data.

gbrammer:
Thanks!

PR #294, opened 2026-06-13
NIRISS trace rotation (FWCPOS)

PJ-Watson:
Adds in a simple trace rotation for TransformGrismconf. Tested with the NGDEEP grismconf-formatted files from [Pirzkal+24](https://github.com/npirzkal/NGDEEP_NIRISS_CALIB), and this matches the rotation already implemented in aXeConf.

The origin of the rotation is computed from the detector location of the 1st order spectrum at a wavelength of 1.047μm, taken from the description in [Willott+22](https://ui.adsabs.harvard.edu/abs/2022PASP..134b5002W/abstract). This wavelength is currently hardcoded, but unless/until the `FWCPOS_REF` keyword is reused for another instrument, there shouldn't be any conflicts.

### DIFF
--- a/grizli/grismconf.py
+++ b/grizli/grismconf.py
@@ -1286,8 +1286,9 @@ class TransformGrismconf(object):
 
     def get_beam_trace_original(self, x=1024, y=1024, dx=0.0, beam="A", fwcpos=None):
         """
-        Function analogous to `grizli.grismconf.aXeConf.get_beam_trace` but
-        that accounts for the different dispersion axes of JWST grisms
+        Function analogous to `grizli.grismconf.aXeConf.get_beam_trace`.
+
+        Calculates the trace in the original (detector) frame of the images.
 
         Parameters
         ----------
@@ -1306,6 +1307,9 @@ class TransformGrismconf(object):
 
         Returns
         -------
+        dx : float or array-like
+            Offset in x pixels from `(x,y)`.
+
         dy : float or array-like
             Center of the trace in y pixels offset from `(x,y)` evaluated at
             `dx`.
@@ -1335,7 +1339,38 @@ class TransformGrismconf(object):
 
         wave = self.conf.DISPL(self.order_names[beam], *x0, t)
 
-        # print(f"t: {t}, delta: {delta}, tdx: {tdx}, tdy: {tdy}, x0[0] + tdx: {x0[0] + tdx}, x0[1] + tdy: {x0[1] + tdy}")
+        # Implement trace rotation due to NIRISS filter wheel position.
+        # Already included in `~grismconf.Config.rotate_trace`, this
+        # uses the grizli sign convention
+        if (fwcpos is not None) and (self.conf.fwcpos_ref is not None):
+
+            # Follow aXeconf convention
+            theta = (fwcpos-self.conf.fwcpos_ref) * np.pi / 180
+
+            if theta != 0.0:
+                # Rotate current trace around the 1st order spectrum,
+                # at a wavelength of 1.047um (Willott+22). Currently
+                # hardcoded, unclear if there are any circumstances
+                # where this is not the correct value
+                origin_t = self.conf.INVDISPL(
+                    self.order_names["A"],
+                    *x0,
+                    self.conf.pivot_wavelength,
+                )
+                origin_x = self.conf.DISPX(self.order_names["A"], *x0, origin_t)
+                origin_y = self.conf.DISPY(self.order_names["A"], *x0, origin_t)
+
+                # rotation = Rotation2D(theta)
+                # tdx_r, tdy_r = rotation(tdx - origin_x, tdy - origin_y)
+                # tdx = origin_x + tdx_r
+                # tdy = origin_y + tdy_r
+
+                tdx, tdy = self.transform.rotate_coordinates(
+                    tdx,
+                    tdy,
+                    theta,
+                    np.squeeze((origin_x, origin_y)),
+                )
 
         return tdx, tdy, wave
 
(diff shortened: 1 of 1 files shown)


TheSkyentist:
Hey @PJ-Watson! @ivastar and I were working on a very similar PR and yours looks great and we'd love to see this merged in grizli.

If you don't mind, we have a few suggestions that we think could help some downstream issues:

1) I think we could move the hardcoded wavelength to a top level variable so its more visible.
2) In `initialize_from_datamodel` we did the following:
`self.pivot_wavelength = getattr(dm, 'pivot_wavelength', 1.05)` as well as 
`self.fwcpos_ref = getattr(dm, 'fwcpos_ref', None)` to track these numbers throughout 

That way a user can overwrite the pipeline default and not worry about the hard coded value if push comes to shove.

We were using 1.05um from an older reference, but I think that these changes could help alleviate any pain down the line in case of alternative pivot wavelengths (i.e. if calibration measurements improve or for the use in other instruments that also have this issue, i.e. Euclid). 

Her full PR to my grizli fork is here: https://github.com/TheSkyentist/grizli/pull/1/changes.

PJ-Watson:
Great minds think alike, it seems! I'd discussed this in an email a few weeks ago with Gaël Noirot, I didn't realise both of you were also working on adding it. I'm relieved that it seems we've got the same solution, I had to double check the sign conventions a few times before I opened this request.

Both those changes make sense to me. The only inconvenience here is that it seems GRISMCONF and specwcs use a slightly different name for the filter wheel reference rotation (capitalised or not), so they have to be unified at some point before calculating the trace.

I've pushed another commit to include those, let me know if that works for your setup as well!

gbrammer:
Thanks for these updates.  A few small points:

- Can you write the keyword arguments on the __init__ method in the old-school Python format to be consistent with the function definitions throughout?  I haven't yet retrained my muscle memory for the `name : type = default` convention in function definitions and I'd rather have the codebase be out-dated but consistent.  What version of python was that convention adopted?  Probably ancient history, I know.
- If the `Rotation2D` object is just a simple rotation matrix multiplication, you can use the [JwstDispersionTransform.rotate_coordinates](https://github.com/gbrammer/grizli/blob/f0062ee4aa1f275d2a2d31ffde21a2d20f88c77b/grizli/grismconf.py#L1080) static method that does the same thing, including specifying the coordinate origin.

gbrammer:
I also just merged in https://github.com/gbrammer/grizli/pull/293, which touches the same function definitions and now raises a conflict here that must be resolved.

PJ-Watson:
Thanks @gbrammer, I meant to keep the original grizli formatting but it looks like that init method slipped through. I'm not sure when the type hinting was originally adopted, but I've been using it since Python 3.10/3.11, when it became possible to write multiple types as e.g. `x : float | np.ndarray`.

Good catch on the matrix rotation, I'm not sure how I'd overlooked that method. I've updated the code to use that instead (with the correct sign convention and units).

Re: #293, I've put the filter wheel rotation inside the `get_beam_trace_original()` function, and updated the docstring accordingly.

ivastar:
FYI, this document says the pivot is at 1.05 um which is what I've been using in the CONF fits:
file:///Users/momcheva/Downloads/Near-_infrared_imager_and_slitless_spectrograph_NI.pdf

ivastar:
@PJ-Watson I think my implementation has the opposite sign in the angle. This is my test:

Screenshot 2026-06-19 at 12 31 50


And this is what the traces look on the actual detector, color-coded by FWCPOS (rotated 180 degrees from above):

Screenshot 2026-06-19 at 12 35 03

PJ-Watson:
@ivastar Thanks for raising this, it's very interesting to see how this rotation appears in the actual data - is the relative scarcity of points at +ve dx caused by fewer programs using the F090W filter?

I think the confusion might be because `Rotation2D` and `JwstDispersionTransform.rotate_coordinates` have the opposite rotation convention. I originally had the same sign as you, but inverted it when swapping the transformation method (see dcab7d5).

I've checked the current commit and can reproduce the same plot you have with the F200W/GR150C combination, so it looks like both our implementations are consistent.
grismconf_test_for_iva

gbrammer:
> @PJ-Watson I think my implementation has the opposite sign in the angle. This is my test:
> 
> Screenshot 2026-06-19 at 12 31 50
> And this is what the traces look on the actual detector, color-coded by FWCPOS (rotated 180 degrees from above):
> 
> Screenshot 2026-06-19 at 12 35 03

Are these offsets relative to some particular config file?  Are you planning to make an updated config?

gbrammer:
Just a comment while you're discussing traces:  way back in the CV ground tests I derived the initial trace calibration assuming a _shared_ (quadratic) curve across all orders and filters.  The last time I checked the CRDS calibrations it seemed like quadratics were being fit to every separate filter + order segment.

There could be a "wedge offset" translation between the filters, but the shape of the trace should be common between them since it's the same dispersing element behind them.  I'm not sure if the orders are expected to share the same trace given the details of the optics, but they seemed to do so in the CV data.

Anyway, deriving a single curve and the spatial dependence of its coefficients should be much more robust than fitting the relatively short segments of each filter / order.

ivastar:
I am in the process of deriving this using a shared curve and the correct pivot point (just trying to finesse the 0th order now). It is true that the current solutions are derived using individual polynomials for each pupil AND also the incorrect pivot point. The whole thing would really only hang together if the derivation of the config and the software implementations are all done consistently.

ivastar:
> @ivastar Thanks for raising this, it's very interesting to see how this rotation appears in the actual data - is the relative scarcity of points at +ve dx caused by fewer programs using the F090W filter?
> 
> I think the confusion might be because `Rotation2D` and `JwstDispersionTransform.rotate_coordinates` have the opposite rotation convention. I originally had the same sign as you, but inverted it when swapping the transformation method (see [dcab7d5](https://github.com/gbrammer/grizli/commit/dcab7d582eee74142967e710c12483cac75f9d87)).
> 
> I've checked the current commit and can reproduce the same plot you have with the F200W/GR150C combination, so it looks like both our implementations are consistent. grismconf_test_for_iva

yes, there are really only 2-3 programs that use the F090W filter (mainly CAL programs) and they don't sample that many values of FWCPOS.

PR #297, opened 2026-06-29
Check for valid pixels in beams

PJ-Watson:
Check that all beams have at least 1 valid pixel when checking for bad PAs in a MultiBeam object. This can occur when fitting beams using a different combination of `min_sens`, `fcontam`, `min_mask`, and `mask_resid` than the beams were extracted with. 

If no valid pixels are present, then the NNLS fitter will crash the python interpreter attempting to fit a zero-dimensional array. This has already been fixed upstream in [`scipy>=1.18`](https://github.com/scipy/scipy/issues/24861), but that version is only 1 week old, and is unlikely to be used in all `grizli` installations for some time.


PJ-Watson:
Sorry, only meant to open a request for the last commit!

PR #301, opened 2026-07-06
Add Option for JWST Pipeline Outlier Rejection/Update to BP masks

alr123:
Added preliminary functionality to run `jwst` pipeline outlier rejection on jwst data with rejection parameters specified in process_direct_grism_visit. This is turned off by default, but can be turned on optionally by setting visit_prep_args['run_jwst_outliers'] = True.  

------------------------------------------------------------------------------------------------------------

Also added new +/- 5*SNMAD outlier "bad" pixels to bad pixel masks. These are created by source-masking and median-stacking background subtracted exposures from two epochs:

1. CEERS (PID 1345) visits 1-4. Many of these outlier pixels may already be flagged in update STScI mask files, but in older observations such as CEERS outlier pixels may not be caught by the older STScI outlier masks. 

2. MINERVA (PID 7814) visits in EGS; observation numbers 25, 26, 27, 28, 29, 30, 31, 32, 33. 

~0.46% new detector pixels are flagged in the NRCALONG chip, and ~0.24% new detector pixels are flagged in the NRCBLONG chip. After merging with the prior most-recent grizli bpmasks, the total percentage of pixels masked are:

- 1.93% and 2.17% in the NRCBLONG and NRCALONG chips, respectively


gbrammer:
Thanks, @alr123 !

I got an exception trying a test with `j072252m7328_smacs0723-glsn-1-12774-f300m_00017`:

```python
jwst_outliers.py:97, in run_jwst_outliers(models, driz_cr_snr, driz_cr_scale)
     88     container.append(m) #add the DataModels made from the grizli-processed rate files into the container.
     90 result = OutlierDetectionStep( #run the rejection step w/ all defualt params except for the snr and scale param
     91     snr=driz_cr_snr,
     92     scale=driz_cr_scale, 
     93     save_intermediate_results=False,       
     94     in_memory=True).process(container)
     96 output_list_of_updated_dq_arrays = [ #make list of output dq arrays that have the new bits set. 
---> 97     np.asarray(result[i].dq, dtype=np.uint32).copy()
     98     for i in range(len(models))
     99 ]
    101 return output_list_of_updated_dq_arrays

TypeError: 'ModelLibrary' object is not subscriptable
```

I'll have a look into it, but you can also try running that association?  Can you also test it on MIRI and NIRISS associations?

And in the meantime, a few additional requests:

1. Add the default `run_jwst_outliers=False` to the default parameters file, perhaps here: https://github.com/gbrammer/grizli/blob/654fbc088d22fc6a352f6ba45d39ccffbce48044/grizli/data/auto_script_defaults.yml#L108
2. Also send the exception traceback to the log after the comment, e.g., `utils.log_exception(utils.LOGFILE, traceback)`.
3. Add docstrings to all functions with the convention like
```python
def get_visit_radec_from_database(visit, nmax=200):
    """
    Get reference astrometry from grizli database

    Parameters
    ----------
    visit : dict
        Visit dictionary, with `product`, `files`, `footprint` keys

    nmax : 200
        Maximum number of alignment sources to return

    Returns
    -------
    radec_file : str
        Creates radec file from the database and returns the filename, or None
        if not overlapping reference sources found
    """
```
4. Add to the docs below `jwst_utils` here: https://github.com/gbrammer/grizli/blob/654fbc088d22fc6a352f6ba45d39ccffbce48044/docs/grizli/index.rst?plain=1#L26
5. It's a minor codestyle thing / habit of mine, but please close the parentheses at the end of a block at the same tab location as the first line of the block. Also generally put comments on their own line above what they refer to, at least for function calls.  Function parameters can have short comments at the end of the line if they don't exceed the PEP8 width of 79 characters (and better still 72). You can try just running `pip install black; black jwst_outliers.py` to standardize the codestyle.  Just send the `black` style with a separate commit from anything that actually changes what the code *does*.

### current
```python
    result = OutlierDetectionStep( #run the rejection step w/ all defualt params except for the snr and scale param
        snr=driz_cr_snr,
        scale=driz_cr_scale, 
        save_intermediate_results=False,       
        in_memory=True).process(container)
```

### new
```python
    # run the rejection step w/ all defualt params except for the snr and scale param
    result = OutlierDetectionStep(
        snr=driz_cr_snr,
        scale=driz_cr_scale, 
        save_intermediate_results=False,       
        in_memory=True
    ).process(container)
```
6. Sorry to be pedantic, but I try to avoid having the `return` statement *do* something as below.  Just break it out into one more additional temporary local variables and return that. 
```python
def hval(header, key, default=""):
    key = key.upper()
    return header.get(ORIGINAL_KEYS.get(key, "O" + key[:7]), header.get(key, default))
```
7. Continuing the pedantry, variable names like `output_list_of_updated_dq_arrays`, are too long for my liking.  I guess PEP8 doesn't have a formal limit other than 79 characters of a full line, which would be silly.  At a minimum, `"output"` is redundant in that example since it's defined and immediately returned ;) .

8. Put `from grizli import jwst_outliers` at the top of the function that uses just below the docstring (imports at the top always, check if needs to be updated other places in your file) and make it rather `from . import jwst_outliers`.  If you're worried about an import exception for a missing jwst dependency, you can wrap it something like
```python
try:
    from . import jwst_outliers
except ImportError:
    jwst_outliers = None
```
and then add it to your test `if (isJWST) and (run_jwst_outliers) and (jwst_outliers is not None):`.

9.  **Important:** I think it would be best to allow *any* modified kwargs to `OutlierDetectionStep` with yet another `visit_prep_args` parameter `jwst_outliers_kwargs={}` (in the function definition in prep.py and in the defaults.yaml file).  To get the behavior using the existing `driz_cr_snr` and scale parameters, in prep.py you can do something like
```python
if "snr" not in jwst_outliers_kwargs:
     jwst_outliers_kwargs["snr"] = driz_cr_snr
```
You'll have to see how to best to pass around `jwst_outliers_kwargs`, but eventually you'd have
```python
def run_jwst_outliers(models, **kwargs):
    [...]
    step = OutlierDetectionStep(**kwargs)
    result = step.process(container)
```
called with `result_dq_arrays = run_jwst_outliers(models, **jwst_outliers_kwargs)`.  I don't think you need to sanitize `**kwargs` to be sure it contains only allowed keywords because `OutlierDetectionStep` itself is defined with `**kws`.

gbrammer:
For debugging that exception, note my environment has 
```
jwst==1.16.1
stpipe==0.7.0
```

alr123:
Hmm that might be the issue. Mine has:
`jwst==1.13.4
stpipe==0.5.1`
And it seems to work there. 

I just tested it on (1) that association you ran, (2) a NIRISS assoc and (3) a MIRI assoc -- and all 3 ran normally in my environment. 

I'll work on those other updates too :)

gbrammer:
The grizli dependency requirement is listed as `jwst<1.17`, so both 1.13.4 and 1.16.1 are valid.  However, 1.16.1 is what you get if you were to do `pip install grizli[jwst]` today, so I'd say best to make it work there.  Both are somewhat out of date compared to the latest major release `jwst==2.0.1` (with `stpipe==1.0.0`), but upgrading grizli + msaexp to use it wasn't immediately trivial so I've tabled that for now.

The `result = OutlierDetectionStep` object I get has an attribute `result._loaded_models` that can be indexed like `result._loaded_models[0].dq`.  I think the original `models[i].dq` of the objects that were put in to the `container` are also updated in place, so perhaps it's safest to get it from there if indeed they include the output from the outlier step.

The code below could be compatible with both 1.13.4 and 1.16.1, but may eventually be incompatible with jwst>=2.0:
```python
    # make list of output dq arrays that have the new bits set.
    try:
        # jwst=1.13
        outlier_dq_arrays = [ 
            np.asarray(result[i].dq, dtype=np.uint32).copy()
            for i in range(len(models))
        ]
    except TypeError:
        # jwst=1.16
        outlier_dq_arrays = [ 
            np.asarray(m.dq, dtype=np.uint32).copy()
            for m in models
        ]
```

gbrammer:
Also **important**:  check which DQ bit `OutlierDetectionStep` is actually flipping.  Grizli assumes that the drizzle-identified bad pixels are DQ=4096, so if `OutlierDetectionStep` sets DQ=4, you can OR the updated DQ array onto the existing array with something like below.
```python

step_outlier_bit = 4   # or whatever it is
grizli_outlier_bit = 4096

with pyfits.open("rate.fits", mode="update") as hdu:
    step_bit_set_ = ((outlier_dq & step_outlier_bit) > 0) > 0
    hdu["DQ"].data |= (step_bit_set_ * grizli_outlier_bit).astype(hdu["DQ"].data.dtype)
    hdu.flush()
```

There's probably a fancier way to do the bit shifting / flipping, but keep it simple and readable. 

And if it *is* DQ=4 that `OutlierDetectionStep` is updating, you should *unset* it before running the outlier step.  DQ=4 is used in Detector1 for any pixel where a jump was identified in the exposure ramps even if the ramp was corrected for the jump.  Deep exposures can have a very high fraction of the pixels with DQ=4, even though the jumps are fixed and the rate images are fine.

Basically, you just want to add any *new* pixels that are flagged by the outlier step, which does seem to be what you're doing after all anyway.  Just make the checks as explicit as necessary.

gbrammer:
One last thing:

I'm inclined to **Merge** the PR now since all of the above discussion is only relevant if you run the new step, which is currently off by default.  That we at least catch the new bpix tables you made in the master branch.

Can you file an [Issue](https://github.com/gbrammer/grizli/issues) that refers to this PR as a reminder that there is work left do be done?

Issue #302, opened 2026-07-07
Updated Required to JWST Outliers module before use


alr123:
Major compatibility, reliability and documentation updates are needed before jwst outliers module, made in PR #301 can be considered safe for general use.

PJ-Watson:
From a compatibility standpoint, I've been running a fork of grizli ([this branch](https://github.com/gbrammer/grizli/compare/master...PJ-Watson:grizli:new_workflow)) with the current `jwst` pipeline for a couple of months now, and haven't encountered any particular problems. These are the relevant package versions in my environment:
```
jwst==2.0.1
stpipe==0.13.0
stcal==1.18.0
stdatamodels==5.0.2
stsci.tools==4.3.0
photutils==2.3.0
drizzlepac==3.11.0
crds==13.2.5
numpy==2.4.6
numba==0.65.1
```
I don't think any of the changes I've made in my fork are related to the pipeline updates, apart from relaxing some of the version requirements. Most of the breaking changes should have been fixed already in the current release of `grizli` - e.g. #286, #288, #291, #292 from pull requests.

PR #303, opened 2026-07-11
Stability/Documentation Update to jwst_outliers module

alr123:
Update to jwst_outliers module. Building on https://github.com/gbrammer/grizli/pull/301, and response to https://github.com/gbrammer/grizli/issues/302

Major changes include:

- stability update so module should now work with `jwst==1.13.x` and also `jwst==1.16.x`, but may be incompatible with `jwst==2.y.x`
- Added option for keyword arguments to be passed to `OutlierDetectionStep`, which may be set in the `prep_args{}` dictionary. Default parameters are set in `.../data/auto_script_defaults.yml`

- Updated flow of `process_direct_grism_visit` so that if `run_jwst_outliers=True`, then the second drizzle rejectoin step using `AstroDrizzle` is skipped. However, If the jwst outlier rejectoin fails for any reason, the `AstroDrizzle` is run instead. The logic for HST associations should be unaffected. 

- Updated the DQ bit which is set by `OutlierDetectionStep` to be DQ=4096, rather than the step defaults. Also updated the logic to clear any pre-existing DQ=4096 or DQ=16 bits before the rejection is run. 

- codestyle and documentation updates.

### DIFF
--- a/grizli/jwst_outliers.py
+++ b/grizli/jwst_outliers.py
@@ -407,7 +407,7 @@ def run_rate_file_group(
 
                 # clear any existing DQ=16 and 4096 bits in file...
                 # like we did in dq model before
-                dq_data = dq_data & CLEAR_BITS
+                dq_data = dq_data & JWST_OUTLIER_CLEAR_MASK
 
                 # set grizli bit where bit=16 was set by this step
                 dq_data |= (
--- a/grizli/jwst_outliers.py
+++ b/grizli/jwst_outliers.py
@@ -315,7 +315,7 @@ def run_rate_file_group(
     rate_files,
     min_files=2,
     verbose=True,
-    jwst_outlier_kwargs=None,
+    jwst_outliers_kwargs=None,
 ):
     """
     Run JWST outlier detection for one group of rate files.
@@ -458,7 +458,7 @@ def do_jwst_outliers_step(
     group_by=("detector", "filter", "pupil", "shape"),
     clean=True,
     verbose=True,
-    jwst_outlier_kwargs=None,
+    jwst_outliers_kwargs=None,
 ):
     """
     Run JWST outlier detection for grouped files in a visit.
@@ -516,7 +516,7 @@ def do_jwst_outliers_step(
             groups[key],
             min_files=min_files,
             verbose=verbose,
-            jwst_outlier_kwargs=jwst_outlier_kwargs,
+            jwst_outliers_kwargs=jwst_outliers_kwargs,
         )
 
     if clean:
--- a/grizli/jwst_outliers.py
+++ b/grizli/jwst_outliers.py
@@ -315,7 +315,7 @@ def run_rate_file_group(
     rate_files,
     min_files=2,
     verbose=True,
-    jwst_outliers_kwargs=None,
+    jwst_outlier_kwargs=None,
 ):
     """
     Run JWST outlier detection for one group of rate files.
@@ -458,7 +458,7 @@ def do_jwst_outliers_step(
     group_by=("detector", "filter", "pupil", "shape"),
     clean=True,
     verbose=True,
-    jwst_outliers_kwargs=None,
+    jwst_outlier_kwargs=None,
 ):
     """
     Run JWST outlier detection for grouped files in a visit.
@@ -516,7 +516,7 @@ def do_jwst_outliers_step(
             groups[key],
             min_files=min_files,
             verbose=verbose,
-            jwst_outliers_kwargs=jwst_outliers_kwargs,
+            jwst_outlier_kwargs=jwst_outlier_kwargs,
         )
 
     if clean:
--- a/grizli/jwst_outliers.py
+++ b/grizli/jwst_outliers.py
@@ -7,9 +7,15 @@ import inspect
 
 import numpy as np
 from astropy.io import fits
+from astropy.stats import sigma_clipped_stats
 
 from grizli import jwst_utils
 from grizli import utils
+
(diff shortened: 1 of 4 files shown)


grizli/jwst_outliers.py
```diff
@@ -1,204 +1,526 @@
...
+                hval(header, "EXP_TYPE", "NRC_IMAGE")
+            ).strip().upper(),
             "shape": tuple(hdul[SCI].data.shape),
         }
-    
+
+    return info
+
+
 def get_mdrizsky(path):
```
PJ-Watson: This function feels slightly unnecessary to me - wouldn't the following work just as well?
```mdrizsky = float(fits.getheader(path, "SCI").get("MDRIZSKY", 0))```
Either way I'd suggest adding a fallback in case the keyword isn't in the expected header.

File: grizli/jwst_outliers.py [hunk 2]
```diff
@@ -1,204 +1,526 @@
...
     with fits.open(path, memmap=False) as hdul:
-        h = hdul["SCI"].header
-        return float(h["MDRIZSKY"])
+        header = hdul[SCI].header
+        mdrizsky = float(header["MDRIZSKY"])
+
+    return mdrizsky
+
 
 def rate_to_cal_path(path):
```
PJ-Watson: A similar comment for this function - I think it'd be better to fold into `run_rate_file_group()`. I'm a fan of using `pathlib.Path` in general in place of strings (and ideally all grizli/jwst/stsci functions would be compatible), but this is only really one line of code to change a filename, and seems to only be used the one time. From my perspective (I might be an outlier on this), it's much easier to read the code when I don't have to jump around to different functions.

grizli/jwst_outliers.py
```diff
@@ -1,204 +1,526 @@
...
+    """
+    utils.log_comment(
+        utils.LOGFILE,
+        "# jwst_rate_outliers: " + str(msg),
+        verbose=verbose,
+        show_date=True,
+    )
+
 
 def pathfix(path):
```
PJ-Watson, line 64: Does the JWST pipeline need absolute paths, or can it handle relative paths?

grizli/jwst_outliers.py
```diff
@@ -1,204 +1,526 @@
...
+
+            log(
+                "updated %s: mapped DQ=%d to DQ=%d; %d outlier pixels flagged. (%.3f%%)"
+                % (
+                    paths[i].name,
+                    JWST_OUTLIER_BIT,
+                    GRIZLI_OUTLIER_BIT,
+                    int(np.count_nonzero(step_outliers)),
+                    float(np.round(100*np.count_nonzero(step_outliers) / npix, 3))
+                ),
```
PJ-Watson:
```suggestion
                f"updated {paths[i].name}: mapped DQ={JWST_OUTLIER_BIT} to DQ={GRIZLI_OUTLIER_BIT}; "
                f"{int(np.count_nonzero(step_outliers))} outlier pixels flagged. "
                f"({np.count_nonzero(step_outliers) / npix:.3%})",
```

PJ-Watson (changes requested):
Just had a quick glance at this - I'm not familiar with the new outliers module, but I wonder if some of the additional functions here could be condensed if they're not utilised elsewhere?

grizli/jwst_outliers.py
alr123: Good point -- will update and also include fallback of sigma-clipped median from the exposure (will require an extra function)

grizli/jwst_outliers.py
alr123: Another good point -- will fold in the actions to `run_rate_file_group()`

grizli/jwst_outliers.py
alr123: Not sure -- but personally I tend to use absolute paths since they work more often than not... This seems to work here so I'm a little hesitant to change unless necessary.

grizli/jwst_outliers.py
alr123: will make this change too

alr123:
Updates to this PR were made -- see https://github.com/alr123/grizli/commit/65f6efa120660f64c41662233fa77fa94576a4bd

grizli/jwst_outliers.py
PJ-Watson: Makes sense, I tend to try for relative paths just in case I need to move directories around - but looking at the rest of the code it seems to be cleaned up within the same function.

PJ-Watson (approved):

grizli/jwst_outliers.py
PJ-Watson: Ah, I meant more in terms of having `.get("MDRIZSKY", 0)`, so the fallback for sky is `0`, rather than `None` and then throwing errors further down the line. But I haven't looked into the details of the outlier detection, so your idea might be more useful.

alr123: I _think_ having it silently default to 0 for the background level is a little more dangerous than having it throw an error. If the background level isn't properly set, then the outlier rejection could potentially flag a whole bunch of pixels that it shouldn't. At any rate, this new method of falling back to estimating the sky level manually (with the sigma-clipped-median) fixes the problem of if an exposure doesn't have MDRIZSKY in it without just setting the sky to 0.

gbrammer:
One last thing: change any `from grizli import []` to `from . import []` in jwst_outliers.py and put them below imports of any other packages.

Otherwise I think it looks good, thanks for the implementation @alr123 and for the discussion @PJ-Watson!

